PR #14, opened 2021-06-06
Add an command to create user

LogischJo:
This PR includes an artisan command to create an admin account mainly for the first installation.

### DIFF
--- a/app/Console/Commands/createUser.php
+++ b/app/Console/Commands/createUser.php
@@ -51,7 +51,7 @@ class createUser extends Command
         if (is_null($response)) {
             print_r('It seems that your Pterodactyl ID isnt correct. Rerun the command and input an correct ID');
 
-            return;
+            return false;
         };
 
         $user = User::create([
@@ -69,5 +69,6 @@ class createUser extends Command
             ['Ptero-ID', $user->pterodactyl_id],
             ['Admin', $user->role],
         ]);
+        return true;
     }
 }
--- a/app/Console/Commands/createUser.php
+++ b/app/Console/Commands/createUser.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use App\Classes\Pterodactyl;
+use App\Models\User;
+use Illuminate\Support\Facades\Hash;
+
+class createUser extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'make:user {--ptero_id=} {--password=}';
+
+    /**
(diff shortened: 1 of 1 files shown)


RamonRobben (approved):
Looks fine to me. @AVMG20 is testing it so he will merge once approved by him.

RamonRobben:
@LogischJo I changed the branch where you are merging to to the Development branch since we will be using that from now on. So for your future pull requests please merge to the Development branch instead of main.

app/Console/Commands/createUser.php
```diff
@@ -0,0 +1,74 @@
...
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {   
+
+        $ptero_id = $this->option('ptero_id') ?? $this->ask('Please specify your Pterodactyl ID.');
+
+        $password = $this->option('password') ?? $this->ask('Please specify your password.');
```
AVMG20, line 46: Can you please check if this password is at least 8 chars :)

app/Console/Commands/createUser.php
```diff
@@ -0,0 +1,74 @@
...
+        
+        $response = Pterodactyl::getUser($ptero_id);
+
+
+        if (is_null($response)) {
+            print_r('It seems that your Pterodactyl ID isnt correct. Rerun the command and input an correct ID');
+
+            return false;
+        };
```
AVMG20, line 55: unnecessary semicolon at the end ;

app/Console/Commands/createUser.php
AVMG20, line 46: Can you please check if this password is at least 8 chars :)

AVMG20 (changes requested):

app/Console/Commands/createUser.php
LogischJo: Do you already have Password Validation rules which I could use?

AVMG20: you only have to verify if the password is at least 8 long for now :) better password validation can be added later on :)

RamonRobben: > Do you already have Password Validation rules which I could use?

As far as I believe we do not currently have this. Password just needs to be at least 8 characters.

LogischJo: k

app/Console/Commands/createUser.php
LogischJo: Removed

app/Console/Commands/createUser.php
LogischJo: Added new Commit

Issue #19, opened 2021-06-07
php artisan update command


LogischJo:
# 🚀 Feature Request

### Is your proposal related to a problem?

Updating the panel requires some manual commands.

### Describe the solution you'd like

Update the panel using an artisan command.

PR #21, opened 2021-06-07
Create the update command

LogischJo:
Create an update command: #19


RamonRobben (approved):
Code looks good to me, I tested it in Docker Environment and the only thing I needed to do was to install Git, after that it ran fine. AVMG is testing it localhost so we'll wait for that. But if you mark this as experimental it will be fine as long as it runs ok of course.

PR #63, opened 2021-06-13
Open links in new tab

anondev-sudo:


resources/views/servers/index.blade.php
```diff
@@ -86,8 +86,8 @@
 
 
                             <div class="card-footer d-flex justify-content-between">
-                                <a href="{{env('PTERODACTYL_URL' , 'http://localhost')}}/server/{{$server->identifier}}" class="btn btn-info mx-3 w-100"><i class="fas fa-tasks mr-2"></i>Manage</a>
-                                <a href="{{env('PHPMYADMIN_URL' , 'http://localhost')}}" class="btn btn-info mx-3 w-100"><i class="fas fa-database mr-2"></i>Database</a>
+                                <a href="{{env('PTERODACTYL_URL' , 'http://localhost')}}/server/{{$server->identifier}}" class="btn btn-info mx-3 w-100" target="__blank"><i class="fas fa-tasks mr-2"></i>Manage</a>
+                                <a href="{{env('PHPMYADMIN_URL' , 'http://localhost')}}" class="btn btn-info mx-3 w-100" target="__blank"><i class="fas fa-database mr-2"></i>Database editor</a>
```
AVMG20, line 90: PHPMyAdmin is not only a database editor but a manager overall. It would be Database Manager instead, but since it's extra, it doesn't really matter, so it doesn't need to be mentioned exclusively.

AVMG20 (changes requested):

resources/views/servers/index.blade.php
anondev-sudo: ok

anondev-sudo: Maybe should be added that when the server product has 0 MySQL databases, then this button would be not displayed

AVMG20: make a new issue for that :)

anondev-sudo: https://github.com/ControlPanel-gg/dashboard/issues/68

anondev-sudo:
I have created new PR https://github.com/ControlPanel-gg/dashboard/pull/70

PR #69, opened 2021-06-16
PaymentController improvents

anondev-sudo:
Adds description, disable that user needs to share their address with owner, and other to paypal payment request

### DIFF
--- a/app/Http/Controllers/Admin/PaymentController.php
+++ b/app/Http/Controllers/Admin/PaymentController.php
@@ -60,7 +60,7 @@ class PaymentController extends Controller
             "purchase_units"      => [
                 [
                     "reference_id" => uniqid(),
-                    "description" => $paypalProduct->display,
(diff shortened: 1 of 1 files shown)


AVMG20:
Has this been tested by you?

anondev-sudo:
I sadly don't have PayPal balance to test it, but I think that it will work. I got it from paypal api docs

RamonRobben:
> I sadly don't have PayPal balance to test it, but I think that it will work. I got it from paypal api docs

You should test it with PayPal Sandbox. If it works in sandbox mode you have a 99% chance that it will work in normal mode aswell. Unless you fill in the wrong API key / secret.

AVMG20:
> I sadly don't have PayPal balance to test it, but I think that it will work. I got it from paypal api docs

You can test it like Ramon says with the Paypal sanbox API tokens, and set your .ENV to local

Issue #72, opened 2021-06-17
Force Discord verification and show alert message


anondev-sudo:
Hi, can you please add config option FORCE_DISCORD_VERIFY in Configuration, if it's set to true, then when the user has verified email, another alert should be shown to verify also Discord account. And user should not be able to create servers and purchase credits when email and discord is not verified.

snaildos:
100% yes.

anondev-sudo:
Another idea: allow users to login using Discord, add Login using Discord button to login page

anondev-sudo:
Maybe there should be that you can't login to the dashboard without verification of email and discord when this is enabled.
That when user register on dashboard, page will be displayed that you must verify your email by clicking on verification link (+ link to resend verification link)  and button to link discord account.

When user who has not verified email and discord logins to the dashboard, this page should be displayed too.

This will require email and discord verification before you can access dashboard. And maybe also without completing this, account on pterodactyl should not be created.

anondev-sudo:
This was added

PR #93, opened 2021-06-25
Forced Discord Account and Mail Verification

LogischJo:
#72 This is configurable using:
VERIFIED_DISCORD_TO_MAKE_SERVER and VERIFIED_EMAIL_TO_MAKE_SERVER

### DIFF
--- a/app/Http/Controllers/ServerController.php
+++ b/app/Http/Controllers/ServerController.php
@@ -11,12 +11,7 @@ use App\Models\Node;
 use App\Models\Product;
 use App\Models\Server;
 use App\Notifications\ServerCreationError;
-use Exception;
-use Illuminate\Contracts\View\Factory;
-use Illuminate\Contracts\View\View;
-use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Auth;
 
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/ServerController.php
```diff
@@ -38,6 +33,16 @@ public function create()
             return redirect()->route('servers.index')->with('error', "You've already reached your server limit!");
         }
 
+        //Required Verification for creating an server
+        if(Configuration::getValueByKey('VERIFIED_EMAIL_TO_MAKE_SERVER') === 'true' && !Auth::user()->hasVerifiedEmail()) {
```
AVMG20, line 37: You can add a default value to 'getValueByKey' by using the second param ('KEY' , 'DEFAULT_VALUE') that way if the key is somehow missing this would still work and not trow errors or unexpected behaviour :)

app/Http/Controllers/ServerController.php
```diff
@@ -38,6 +33,16 @@ public function create()
...
+            return redirect()->route('profile.index')->with('error', "You havent verified your email! Thats required to create an server.");
+        }
+
+        //Required Verification for creating an server
+        if(Configuration::getValueByKey('VERIFIED_DISCORD_TO_MAKE_SERVER') === 'true' && !Auth::user()->discordUser) {
```
AVMG20, line 42: check comment above :)
To what method have you added this script? should be added to both the create and store method to prevent any bypasses

database/seeders/ConfigurationSeeder.php
```diff
@@ -19,52 +19,66 @@ public function run()
...
+            'key' => 'VERIFIED_EMAIL_TO_MAKE_SERVER',
+        ], [
+            'value' => 'true',
+            'type' => 'boolean',
+        ]);
+
+        Configuration::firstOrCreate([
+            'key' => 'VERIFIED_DISCORD_TO_MAKE_SERVER',
+        ], [
+            'value' => 'true',
```
AVMG20, line 50: not much besides turning on the extra spacing options for array's :P 
could you also rename the keys to something like (FORCE_EMAIL_VERIFICATION)
and perhaps add a small description explaining what this option does :)
I've already done this to all of the config options, but it seems like you are running an older version.

AVMG20 (changes requested):

app/Http/Controllers/ServerController.php
LogischJo: Done in new PR (#95) (sadly I was not able to commit into this one)

app/Http/Controllers/ServerController.php
LogischJo: Done in new PR (#95) (sadly I was not able to commit into this one)

database/seeders/ConfigurationSeeder.php
LogischJo: Done in new PR (#95) (sadly I was not able to commit into this one)

AVMG20:
Already solved

PR #101, opened 2021-07-01
Email formatting

anondev-sudo:
Added spaces to confirmed payment email, as some mail clients cannot format it well
image

### DIFF
--- a/resources/views/mail/payment/confirmed.blade.php
+++ b/resources/views/mail/payment/confirmed.blade.php
@@ -1,16 +1,16 @@
 @component('mail::message')
 # Thank you for your purchase!
-Your payment has been confirmed; Your credit balance has been updated.
+Your payment has been confirmed; Your credit balance has been updated.<br>
 
 # Details
 ___
-### Payment ID: **{{$payment->id}}**
-### Status:     **{{$payment->status}}**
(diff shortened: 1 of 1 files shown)


anondev-sudo:
Will be this merged?

RamonRobben:
@anondev-sudo can you attach a screenshot of how the mail looks like after your changes? I am also thinking like if its a list of items then why don't we make it an html list but then without the dots. Newlines are good but if its a list its probably easier to make a list instead of newlines.

anondev-sudo:
Yes, it could be in html instead of markdown

anondev-sudo:
It will look normal on that email client.
But when I have tried to view the original email on gmail by forwarding it, it looked normal, so the problem is just for some email clients

RamonRobben:
@anondev-sudo if instead of text-formatting you use html tags it should work on almost all email clients.

AVMG20:
> @anondev-sudo if instead of text-formatting you use html tags it should work on almost all email clients.

Laravel converts the markdown to html format

RamonRobben:
@AVMG20 ah okay I didn't know that. We can still use HTML in the template tho.
Gives users the extra ability to write some html themselves or write inline css to change the look even more.
And more customization is always better.

PR #149, opened 2021-07-22
ControlPanel CopyRight & AddCredits endpoint

Hiekki4:


routes/api.php
```diff
@@ -17,6 +17,7 @@
 */
 
 Route::middleware('api.token')->group(function () {
+    Route::patch('/users/{server}/addcredits', [UserController::class, 'addCredits']);
```
AVMG20, line 20: Instead of {server} add {user}

app/Http/Controllers/Api/UserController.php
```diff
@@ -63,6 +63,27 @@ public function update(Request $request, int $id)
...
+    public function addCredits(Request $request, int $id)
+    {
+        $discordUser = DiscordUser::find($id);
+        $user = $discordUser ? $discordUser->user : User::findOrFail($id);
+
+        $request->validate([
+            "credits"      => "required|numeric|min:0|max:1000000",
+        ]);
+
+        $user->increment('credits', $request->credits);
```
AVMG20, line 82: Make sure to check if the users credits doesn't go over the max amount of credits before adding it. If it does go over the max ammount this wil result in a 500 sql error.

Example check pulled from voucher controller (redeem function)
```
if ($request->user()->credits + $voucher->credits >= 99999999) throw ValidationException::withMessages([
            'code' => "You can't redeem this voucher because you would exceed the credit limit"
        ]);
```

AVMG20 (changes requested):

PR #152, opened 2021-07-27
increment & decrement endpoints

Hiekki4:


AVMG20 (changes requested):
validation on decrement incorrect

PR #164, opened 2021-08-08
Include discord_id parameter in App\Models\User.php

DarkGuy10:


AVMG20:
Can you describe why you want this changed? Currently, our update code isn't being used yet to update discord links.

PR #165, opened 2021-08-09
Fix allocation limit

j122j:
Use product's allocation limit instead of 1.

### DIFF
--- a/app/Classes/Pterodactyl.php
+++ b/app/Classes/Pterodactyl.php
@@ -33,7 +33,8 @@ class Pterodactyl
      * @param int $pterodactylId
      * @return mixed
      */
-    public function getUser(int $pterodactylId){
+    public function getUser(int $pterodactylId)
+    {
         $response = self::client()->get("/application/users/{$pterodactylId}");
 
         if ($response->failed()) return $response->json();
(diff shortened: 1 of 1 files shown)


ok236449:
This issue was created 26 days ago and it still wasn't fixed in the "prerelease" version. Propably only in development. It shouldn't be hard, it is propably just a typo because all the other variables are used correctly when ordering a server. Why is there such a big delay?

RamonRobben:
> This issue was created 26 days ago and it still wasn't fixed in the "prerelease" version. Propably only in development. It shouldn't be hard, it is propably just a typo because all the other variables are used correctly when ordering a server. Why is there such a big delay?

I mean the fix has been merged to Development. There just hasn't been a release in a while due to various circumstances.
The goal is to have a stable development branch before creating a release.

ok236449:
Ok then. And do you think the release will come out let's say in the next month? Any idea if it will be a full release, or a prerelease something like v0.5?

RamonRobben:
@ok236449 Yeah should be this week maybe. Depends on how much time the original creator has to test all the merges etc.
Yeah other payment methods, and server add-ons are the more major additions that are not yet planned. Mainly because everyone has a different opinion on payment method / provider. I think Stripe was in the planning but I am unsure wether it was scrapped or not. For add-ons it has not yet been planned since its a quite big addition.

PR #195, opened 2021-09-20
Disabled button after button push.

nickdevnl:
I disabled the submit button after you push it.
So you can't spam click the button and create 30 servers and you can only create one a time.
If you have more question dm me on discord ItsJustNickDev#9943


anondev-sudo:
Good idea

PR #198, opened 2021-09-22
Show Server price per Hour / per Month

1day2die:
Comming from Issue :  https://github.com/ControlPanel-gg/dashboard/issues/196

### DIFF
--- a/resources/views/servers/index.blade.php
+++ b/resources/views/servers/index.blade.php
@@ -84,8 +84,12 @@
                                         <td>{{$server->product->backups}}</td>
                                     </tr>
                                     <tr>
-                                    <td>Price per Hour: {{number_format($server->product->getHourlyPrice(),2,".")}} Credits</td>
-                                    <td>Price per Month: {{$server->product->getHourlyPrice()*24*30}} Credits</td>
+                                        <td>Price per Hour</td>
+                                        <td>{{number_format($server->product->getHourlyPrice(),2,".")}} Credits</td>
+                                    </tr>
+                                    <tr>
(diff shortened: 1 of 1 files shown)


1day2die:
image

1day2die:
> Please use proper indentation.
> Also I would keep the structure of the table the same so title left (in this case Price per month or hour) and on the right the value.
> 
> So like:
> Price per hour: 100 credits
> Price per month: 3750 credits

Ah okay.
I did it like this because the card would become so long otherwise

j122j:
What if the hourly credit usage was here?
Right/below in the same box, or in the separate box on the right, there is space there.

image

1day2die:
> What if the hourly credit usage was here?
> Right/below in the same box, or in the separate box on the right, there is space there.
> 
> image

That space is already used by this Pull Request

https://github.com/ControlPanel-gg/dashboard/pull/197


See

anondev-sudo:
Good idea

anondev-sudo:
Also there is configuration credit display name, so you should use it

anondev-sudo:
Also why is server description not displayed in server card?

1day2die:
> Also there is configuration credit display name, so you should use it

Oh snap, of course.

>Also why is server description not displayed in server card?

dunno. it never was

anondev-sudo:
You can add description table row too

AVMG20:
looks good, we do have plans on changing the server listing soon. so it might not even see daylight in main branch (not sure when si i merge it ) Good job, thanks for both pull request 👍

PR #199, opened 2021-09-25
Fix division by zero

j122j:
The page shows error 500 if usage is 0, this pull request removes the Out of Credits info box if the usage is 0.

### DIFF
--- a/resources/views/home.blade.php
+++ b/resources/views/home.blade.php
@@ -67,6 +67,7 @@
                 </div>
                 <!-- /.col -->
 
+                @if($useage > 0)
                 <div class="col-12 col-sm-6 col-md-3">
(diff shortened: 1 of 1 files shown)


anondev-sudo:
Maybe to write there already out of credits or so

anondev-sudo:
Why you dont write there that you ran out of credits already

j122j:
> Why you dont write there that you ran out of credits already

Because you didn't run out of credits, you just weren't using any.
I had 900 credits but no servers, which means my usage will be 0 and the home page will error.

anondev-sudo:
Ok, what will be showed when you ran out of credits?

j122j:
#201

Issue #204, opened 2021-09-26
Suspend user account


anondev-sudo:
# 🚀 Feature Request

### Is your proposal related to a problem?

*Provide a clear and concise description of what the problem is.
For example, "I can't change x and it makes me have to do manual work"*

Recently our client purchased credits, but the payment has failed. The client got the credits added to the balance. 5 days after the purchase have received an email from PayPal that the payment was not successful.
![image](https://user-images.githubusercontent.com/85408287/134803671-b2758fd0-31c8-45da-811f-8a1c7f9010bd.png)
We asked the client to repeat the payment manually, and we have suspended their servers for the time we get it.
I got an idea that there should be an option to suspend a user account on the controlpanel.

### Describe the solution you'd like

*Provide a clear and concise description of what you want to happen.
For Example, "Automate the changing of x so I don't have to do it manually"*

Add button Suspend user to admin area on the user page. When a user is suspended, all their servers should be suspended automatically and when the user wants to access the controlpanel, an error message should be displayed like `Your account was suspended, please contact support for more information.`.  User should not be able to do anything on the dashboard, including purchasing credits and editing profile. Maybe when the user is suspended, the dashboard could automatically logout them and display an error message.

When a user is suspended, in the admin area there should be the button Unsuspend user, when you click on it, the user should be unsuspended and all their servers should be unsuspended too.

### Additional context

*Is there anything else you can add about the proposal?
You might want to link to related issues here, if you haven't already.*

1day2die:
PullRequest created

PR #205, opened 2021-09-26
Add the ability to Suspend Users

LogischJo:
This allows Admins to suspend User Accounts. An suspended account gets logged out in the next request and an information message gets displayed (on the login page). 
Refers to #204

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -59,7 +59,8 @@ class User extends Authenticatable implements MustVerifyEmail
         'password',
         'pterodactyl_id',
         'discord_verified_at',
-        'avatar'
+        'avatar',
+        'suspended'
     ];
 
     /**
@@ -79,7 +80,7 @@ class User extends Authenticatable implements MustVerifyEmail
      */
     protected $casts = [
         'email_verified_at' => 'datetime',
-        'last_seen' => 'datetime',
+        'last_seen'         => 'datetime',
     ];
 
     /**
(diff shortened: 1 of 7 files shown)


LogischJo:
It is inspired by the suspension system of servers because you don't have to reinvent the system every time 😛

LogischJo:
Here some screenshots:

anondev-sudo:
Good

anondev-sudo:
Can you contribute also on referral system or egg/node linking to products if you have times?

LogischJo:
> Can you contribute also on referral system or egg/node linking to products if you have times?

This PR is dedicated to User Suspension, but am I working with AVMG and others on the big new linking feature for Products with nodes and eggs.. Some news about this soon hopefully. The referral feature is an cool idea too but as of now it is not planned by me, but who knows if I have time for it soon

LogischJo:
AVMG and the other devs just reported me back that it would make more sense to suspend all the current running servers from an user.  I really forgotten that, I will look into it tomorrow and then this should be ready to get merged

1day2die:
working fine

app/Models/User.php
```diff
@@ -189,6 +193,12 @@ public function suspend()
      */
     public function unSuspend()
     {
+        foreach ($this->servers as $server){
+          if ($this->credits >= $server->product->getHourlyPrice()) {
+            $server->unSuspend();
+        }
```
BitSec01, line 199: Make sure the formatting is correct when pushing a PR to keep things clean and have the brackets aligned.
Seems like you need to select line 198 and 199 and press tab and it should be fixed.

LogischJo: Got fixed in the new commit 👍

AVMG20 (approved):

BolverBlitz:
If I see that correctly, it also includes an API call?

PR #207, opened 2021-09-26
Replace pterodactyl link with dashboard link

anondev-sudo:
Replace pterodactyl link with dashboard link in table and add pterodactyl link on show user page

### DIFF
--- a/app/Http/Controllers/Admin/UserController.php
+++ b/app/Http/Controllers/Admin/UserController.php
@@ -282,7 +282,7 @@ class UserController extends Controller
                 return '<span class="badge ' . $badgeColor . '">' . $user->role . '</span>';
             })
             ->editColumn('name', function (User $user) {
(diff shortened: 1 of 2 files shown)


AVMG20:
Uhm, Why? This button already brings you to our show page. Now the name and button bring you to our show? I don't understand why you want to make this change. I think it's very convenient to have the ability to go to the show page of either ptero or dashboard while browsing on the index page.
image

anondev-sudo:
Because it's easier when I click on the link.

There is problem that I can't use ID or email from pterodactyl in API request. So when I click on the user name, I get pterodactyl ID, but on pterodactyl there is not dashboard id. 

Option to use pterodactyl and dashboard IDs in API will solve my issue.

AVMG20:
> Because it's easier when I click on the link.
> 
> There is problem that I can't use ID or email from pterodactyl in API request. So when I click on the user name, I get pterodactyl ID, but on pterodactyl there is not dashboard id.
> 
> Option to use pterodactyl and dashboard IDs in API will solve my issue.

You can click on the eye button, see the user profile page and copy the Ptero ID from there. this has always been there and changing the link does have nothing to do with this issue besides removing useful functionality.

anondev-sudo:
Ok

PR #210, opened 2021-09-27
Added "ran out of Credits"

1day2die:
image

### DIFF
--- a/resources/views/home.blade.php
+++ b/resources/views/home.blade.php
@@ -65,8 +65,9 @@
                     </div>
                     <!-- /.info-box -->
                 </div>
-                <!-- /.col -->
 
+                <!-- /.col -->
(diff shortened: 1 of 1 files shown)


AVMG20:
I think it's better to hide the box entirely when you have 0 credits, simply because not every host gives away free credits. And now these hosts get clients that 'ran out of credits' even tho they never had credits, to begin with.

1day2die:
> I think it's better to hide the box entirely when you have 0 credits, simply because not every host gives away free credits. And now these hosts get clients that 'ran out of credits' even tho they never had credits, to begin with.

The Box will now be hidden, if you are a new user with 0 Credits and no running servers.
It will also be hidden, if you got credits but got no running servers

It will be shown, if you got credits and running servers
it will show "Ran out of Credits" for the last Hour your server is online


This way we got everything covered!

shawn3308:
hello guys xD

PR #211, opened 2021-10-01
Add minimal credits per product

### LINKED ISSUE #161: Minum Credits per Product[feature]
# 🚀 Feature Request

First of all, i am pretty sure that i have already seen that Feature Request Somewhere, so sorry if this is a Duplication, i couldn't find it right now...

### Is your proposal related to a problem?

*Provide a clear and concise description of what the problem is.
For example, "I can't change x and it makes me have to do manual work"*

Lets say i have 20 Credits and i have a plan for 1 Credit / Hour. I could totally buy that.. But what happens if i have a plan for 50 Credits / Hour? Maybe something more Premium? Then the user could Create the Expensive server (because he doesn't need to pay at creation) can use the server for one hour and than the server suspends. So he had like one hour of free server.

Of couse you could fix that with setting the Minium Credits to 50 but then he could not buy the Product which costs 1 Credit / Hour

### Describe the solution you'd like

*Provide a clear and concise description of what you want to happen.
For Example, "Automate the changing of x so I don't have to do it manually"*

An Option for Every Product to set the Minimum Credits needed to create an server :)

### Additional context

*Is there anything else you can add about the proposal?
You might want to link to related issues here, if you haven't already.*

I discussed on that issue with @${j122j} on discord :)

j122j:
This pull request resolves #161.

~~This is currently a draft and hasn't been tested.~~

### DIFF
--- a/app/Http/Controllers/Admin/ProductController.php
+++ b/app/Http/Controllers/Admin/ProductController.php
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers\Admin;
 
 use App\Http\Controllers\Controller;
+use App\Models\Configuration;
 use App\Models\Product;
 use Exception;
 use Illuminate\Contracts\Foundation\Application;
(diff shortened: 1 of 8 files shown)


ZombieBro2044:
Merge when really need this feature 😊.

j122j:
Screenshots
image
image

LogischJo:
Yoo, I just tried this and its so cool. But I guess it would be better if we display an error using an sweetalert if he doesn't have enough balance

PR #227, opened 2021-10-13
If discord is connected, use discord avatar instead of gravatar

1day2die:
https://github.com/ControlPanel-gg/dashboard/pull/226

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -213,11 +213,12 @@ class User extends Authenticatable implements MustVerifyEmail
      */
     public function getAvatar()
     {
-        if ($this->discordUser()->exists()){
+        if ($this->discordUser()->exists()) {
             $avatar = $this->discordUser->getAvatar();
-        }else{
+        } else {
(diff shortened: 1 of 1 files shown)


anondev-sudo:
It's not working, I still see gravatar default avatars when viewing users in admin area.
Also users see their gravatars.
image

j122j:
This was removed in https://github.com/ControlPanel-gg/dashboard/commit/deb4e897014102b9eb7ef19d801ad27ca70972ee

anondev-sudo:
Oh, should be added back

1day2die:
no. it increased loading times by a lot. you can grab the code from here and use in your panel though

https://github.com/ControlPanel-gg/dashboard/pull/226/files

SneakyHub:
You can literally cache the discord avatars when the dashboard syncs statistics and use the cached avatars instead of live avatars that way loading times are basically the same,

IceToast:
Exactly thats meant with "other ways" :D in the mentioned issue

PR #247, opened 2021-10-28
Sweetalert confirmation before deleting server

anondev-sudo:
image

When you cancel:
image


AVMG20:
This pr is very hardcoded. Doing it this way, you would have to copy-paste this code on every place you wish to implement this delete pop-up. I would love to see a more global solution by making a function/method which can be easily added to the places you want this pop-up instead.

anondev-sudo:
So should I make js function confirmSubmit and then add property `onsubmit="confirmSubmit(this)"` to the form?

anondev-sudo:
I don't see problem in it that it's hardcoded

PR #265, opened 2021-11-05
Full configurable Tax support

1day2die:
**`sudo php artisan migrate --seed --force` must be executed to test this PR**

### Full Tax support using Configurations to set the Tax

image

image

### Splitting Tax and Productprice on Paypal
image

### and full payment-log support

image


anondev-sudo (approved):
Nice

app/Models/Payment.php
```diff
@@ -53,10 +53,9 @@ public function user()
     {
         return $this->belongsTo(User::class);
     }
-
-    public function formatCurrency($locale = 'en_US')
+    public function formatToCurrency($value,$locale = 'de_DE')
```
LogischJo, line 56: Please change the fallback back to en_US

LogischJo:
Not bad, i will debug this PR on my machine and then approve your requests

app/Models/Payment.php
1day2die: we're working with €-Currency as default. 
You write the €-sign behind the money-amount -> 10,00€ and not €10,00

I  dont see why it should default to US, basically the wrong way around

BitSec01: Perhaps we can implement some currency formatting for all countries? Not sure how hard it is to implement but for the proper formatting you can look at Shopify: https://polaris.shopify.com/foundations/formatting-localized-currency

1day2die: its not needed i guess, its not bad having it aswell though.

 it just was there and i did not think about it tbh, just reused it

j122j:
Maybe getTaxValue and getTotalPrice should be moved to PaypalProduct and not PaymentController. This way you can just pass product to the view and use $product->getTotalPrice(). Also in formatCurrency you could do $product->getTotalPrice().

1day2die:
> Maybe getTaxValue and getTotalPrice should be moved to PaypalProduct and not PaymentController. This way you can just pass product to the view and use $product->getTotalPrice(). Also in formatCurrency you could do $product->getTotalPrice().

oh yeah thats obvious. will do, thanks bud

1day2die:
> Maybe getTaxValue and getTotalPrice should be moved to PaypalProduct and not PaymentController. This way you can just pass product to the view and use $product->getTotalPrice(). Also in formatCurrency you could do $product->getTotalPrice().

Done. Thanks bud!

app/Models/Payment.php
AVMG20: Default 'de_DE'?

This method was to format currency for the payment price. You're no longer using 'this' object but instead made this a global function that no longer belongs in this model since it could be used by anything now.

Usually, I would ask you to move it to some helper class instead, but looking at the code, I think it's okay ;) 

+ missing docblocks

app/Models/PaypalProduct.php
```diff
@@ -43,9 +44,25 @@ public static function boot()
...
      */
-    public function formatCurrency($locale = 'en_US')
+    public function formatToCurrency($value,$locale = 'de_DE')
     {
         $formatter = new NumberFormatter($locale, NumberFormatter::CURRENCY);
-        return $formatter->formatCurrency($this->price, $this->currency_code);
+        return $formatter->formatCurrency($value, $this->currency_code);
+    }
+
+    public function getTaxPercent(){
```
AVMG20, line 53: No docblocks

app/Models/PaypalProduct.php
```diff
@@ -43,9 +44,25 @@ public static function boot()
...
+        $tax = Configuration::getValueByKey("SALES_TAX");
+        if ( $tax < 0 ) {
```
AVMG20, line 55: can be a shorthand if statement
` return $tax < 0 ? 0 : $tax;`

app/Models/PaypalProduct.php
```diff
@@ -43,9 +44,25 @@ public static function boot()
...
+            return 0;
+        }
+        return $tax;
+    }
+
+    public function getTaxValue(){
```
AVMG20, line 61: no docblocks
add some spaces in the method to make it a bit better readable

app/Models/PaypalProduct.php
```diff
@@ -43,9 +44,25 @@ public static function boot()
...
+        return $this->price*$this->getTaxPercent()/100;
+    }
+
+    public function getTotalPrice(){
```
AVMG20, line 65: no docblocks
add some spaces in the method to make it a bit better readable

AVMG20 (changes requested):

j122j (approved):

app/Models/Payment.php
```diff
@@ -51,9 +54,14 @@ public function user()
         return $this->belongsTo(User::class);
     }
 
-    public function formatCurrency($locale = 'en_US')
+    /**
+     * @param  float/int
```
AVMG20, line 58: invalid docblock
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md

app/Models/PaypalProduct.php
```diff
@@ -40,12 +41,41 @@ public static function boot()
     }
 
     /**
+     * @param  float/int
      * @param string $locale
      * @return string
      */
-    public function formatCurrency($locale = 'en_US')
+    public function formatToCurrency($value,$locale = 'en_US')
```
AVMG20, line 48: invalid docblock

app/Models/PaypalProduct.php
```diff
@@ -40,12 +41,41 @@ public static function boot()
...
         $formatter = new NumberFormatter($locale, NumberFormatter::CURRENCY);
-        return $formatter->formatCurrency($this->price, $this->currency_code);
+        return $formatter->formatCurrency($value, $this->currency_code);
+    }
+
+    /**
+     * @desc Returns the tax in % or 0 if less than 0
+     * @return int
+     */
+    public function getTaxPercent()
```
AVMG20, line 58: invalid docblock

app/Models/PaypalProduct.php
```diff
@@ -40,12 +41,41 @@ public static function boot()
...
+    {
+        $tax = Configuration::getValueByKey("SALES_TAX");
+        return $tax < 0 ? 0 : $tax;
+    }
+
+    /**
+     * @desc Returns the total tax value.
+     * @return float
+     */
+    public function getTaxValue()
```
AVMG20, line 68: invalid docblock

app/Models/PaypalProduct.php
```diff
@@ -40,12 +41,41 @@ public static function boot()
...
+    {
+        return $this->price*$this->getTaxPercent()/100;
+    }
+
+    /**
+     * @desc Returns the total price incl. tax
+     * @return float
+     */
+    public function getTotalPrice()
```
AVMG20, line 77: invalid docblock

AVMG20 (changes requested):

PR #275, opened 2021-11-08
Fix "You Ran out of Credits - hours"

1day2die:


AVMG20:
what's the fix for what? unclear title / description

app/Http/Controllers/HomeController.php
```diff
@@ -48,6 +48,30 @@ public function getTimeLeftBoxBackground(float $days)
...
+    *
+    * @return string
+    */
+    public function getTimeLeftBoxUnit(float $days, float $hours)
+    {
+        if ($days < 1)
+        {
+            if ($hours < 1)
+            {
+                return;
```
AVMG20, line 65: don't return void if your return docblock says string
can also return null and make return type !string
this code can be simplified

1day2die:
> what's the fix for what? unclear title / description

please check the linked issue.

app/Http/Controllers/HomeController.php
```diff
@@ -89,7 +113,7 @@ public function index(Request $request)
 
             $bg = $this->getTimeLeftBoxBackground($days);
             $boxText = $this->getTimeLeftBoxText($days, $hours);
-            $unit = $days < 1 ? 'hours' : 'days';
+            $unit = $this->getTimeLeftBoxUnit($days, $hours);
```
AVMG20, line 116: shorthand = return $days < 1 ? ($hours < 1 ? null : "hours") : "days";

AVMG20 (changes requested):

PR #276, opened 2021-11-08
Show Client running out of Credits in Admin overview

1day2die:
Dont know if this is needed, but its kinda a "funny to know" piece of information and there was an Issue open for this matter

image

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -242,6 +242,12 @@ class User extends Authenticatable implements MustVerifyEmail
         return number_format($usage, 2, '.', '');
     }
 
+
+    /**
+    * @description Returns the Users "out of Credits" time
+    *
+    * @return string
+    */
     public function outOfCredits()
     {
         $usage = $this->creditUsage();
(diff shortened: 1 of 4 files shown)


anondev-sudo:
When user ran out of credits, should be displayed

app/Models/User.php
```diff
@@ -242,6 +242,26 @@ public function creditUsage()
         return number_format($usage, 2, '.', '');
     }
 
+
+    /**
+    * @description Returns the Users "out of Credits" time
+    *
+    * @return string
+    */
+    public function outOfCredits()
```
AVMG20, line 251: purchase phpstorm or install a beutify plugin :)

AVMG20 (changes requested):

app/Models/User.php
1day2die: DocBlock or Formatting ?

app/Models/User.php
```diff
@@ -242,6 +242,28 @@ public function creditUsage()
         return number_format($usage, 2, '.', '');
     }
 
+
+    /**
+    * @description Returns the Users "out of Credits" time
+    *
+    * @return string
+    */
+     * @return string
```
AVMG20, line 251: what's going on here

AVMG20 (changes requested):

anondev-sudo:
I don't see it
image

PR #295, opened 2021-11-13
Just some visual ones

anondev-sudo:


anondev-sudo:
image
To prevent from pressing it more times ...

nickdevnl (changes requested):
This change doesnt work now there are 2 create buttons

nickdevnl (approved):
jk

IceToast:
Reworked the whole create server UI. #315 
This PR would be obsolete then

1day2die:
Thanks for this anon, however as IceToast stated already this has been reworked a whole lot :)

anondev-sudo:
> image To prevent from pressing it more times ...

Make sure to add there this too

1day2die:
@IceToast  hes got a point

Issue #294, opened 2021-11-13
Display server egg, nest, location, node and product name on the server's card.


anondev-sudo:
# 🚀 Feature Request

### Is your proposal related to a problem?

*Provide a clear and concise description of what the problem is.
For example, "I can't change x and it makes me have to do manual work"*

User's don't know what plan are they using on the server, what egg they have there and so.

### Describe the solution you'd like

*Provide a clear and concise description of what you want to happen.
For Example, "Automate the changing of x so I don't have to do it manually"*

Display more details about the server like egg, nest, location, node and product name.

Also make that resources (RAM, CPU) will be displayed inside one table row. Should be similiar to card that is displayed when you are making new server.

![image](https://user-images.githubusercontent.com/85408287/141643144-eec00596-ee61-4aa7-ab00-6b5fdc5e8fdb.png)

Details about servers from pterodactyl should be fetched daily and cached. And then cached data should be used there. 
*This will also solve bug that when you rename server on pterodactyl, name is not updated on dashboard.*
And when you transfer the server to different node, or change egg, it would be updated after hour.
Cronjob should handle this. And also when server got removed from pterodactyl (404 status was returned), delete the server from database also on dashboard.

### Additional context

*Is there anything else you can add about the proposal?
You might want to link to related issues here, if you haven't already.*

(Write your answer here.)

1day2die:
will be closed by 
https://github.com/ControlPanel-gg/dashboard/pull/315

anondev-sudo:
This is meant on the servers pages. Not on servers create page

anondev-sudo:
The PR does not fix this issue so unlink it

1day2die:
PR open

PR #315, opened 2021-11-27
servers/create - updated UI

IceToast:
Updated the servers/create page to a more mordern UI. The ressource selection is now a list of cards to select from.

Screenshot of UI:
image


1day2die:
Hot

anondev-sudo:
Update also page that will see userd when server is created (Servers tab in the menu) to similar cards

IceToast:
> Update also page that will see users when server is created (Servers tab in the menu) to similar cards

That's a whole different topic and has nothing to do with this pr.
This pr is focused on server creation only

anondev-sudo:
The design should be the same. Also how will server creation page look when you have more than 5 plans, will there be max 5 plans in a row? And how does it look on phone?

IceToast:
The design do not have to be the same as they serve a much different need of information. If you want it to be same, go for it.

Why would there be a limit. They would overflow to the next row. 

This project is not mobile focused at all, but using grids like that will make the whole app very responsive. On mobile it would be exactly the same but not as many cards in a row obviously...

anondev-sudo:
Like it, but can you make similiar cards with available information for servers page? Like the fontawesome icons, smaller cards

anondev-sudo:
Also it would be nice to include node, location, egg, nest on the card too. On the original page it is.
image

IceToast:
Dude I told you already, that is not the point of this PR. make an issue or make it yourself!

anondev-sudo:
Why removing things that were there before?

IceToast:
> Why removing things that were there before?

What the hell are you talking?! I redesigned the UI. Stop demanding me to do anything you want to have, do it yourself.

There is not information being removed. It's just way more modern.

Now stop commenting on this PR!

anondev-sudo:
So where is chose egg, node and nest on the card? You removed it

IceToast:
> So where is chose egg, node and nest on the card? You removed it

Are you stupid? Its right there, just like before. Node, Software/Game (aka. Nest) and Specification (aka. Egg)...

anondev-sudo (commented):
It was removed. Selected nest, egg and node is not showed on the product cards. You can see it just in the select form elements.

IceToast:
> It was removed. Selected nest, egg and node is not showed on the product cards. You can see it just in the select form elements.

What you are talking about was the "checkout summary" on the right side, which gave a summary of the whole configuration, to check before buying. 

This PR is about the "resource plan"-selection, which i redesigned to be a modern card selection. 
As you can tell (from the screenshot above) you can see all the configuration information without the need of a summary.
I removed the summary on the right side, as it breaks the centered design of this redesign.

I tried rearranging the UI components, but that didn't help. Thinking about why the summary was there in the first place i decided to remove it as it does not serve any needs. All the information is visible at one glance.

anondev-sudo:
#294 is not implemented in this PR, this includes just redesign of server create page, servers page is not redesigned

IceToast:
> #294 is not implemented in this PR, this includes just redesign of server create page, servers page is not redesigned

This comment is not related to this PR but the issue you are talking about, please keep it to the mentioned issue

PR #317, opened 2021-11-29
fix: 🚑️ swapped nodes/eggs column naming -> now correct order

IceToast:
Fixed the column naming (nodes & eggs) in the products data table

image

### DIFF
--- a/resources/views/admin/products/index.blade.php
+++ b/resources/views/admin/products/index.blade.php
@@ -11,8 +11,7 @@
                 <div class="col-sm-6">
                     <ol class="breadcrumb float-sm-right">
                         <li class="breadcrumb-item"><a href="{{ route('home') }}">Dashboard</a></li>
-                        <li class="breadcrumb-item"><a class="text-muted"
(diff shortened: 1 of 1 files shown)


1day2die (approved):

resources/views/admin/products/index.blade.php
```diff
@@ -74,30 +74,63 @@ function submitResult() {
...
-                    {data: 'disk'},
-                    {data: 'databases'},
-                    {data: 'backups'},
-                    {data: 'nodes', sortable: false},
-                    {data: 'eggs', sortable: false},
-                    {data: 'servers', sortable: false},
-                    {data: 'created_at'},
-                    {data: 'actions', sortable: false},
+                order: [
+                    [2, "asc"]
```
1day2die, line 83: Please bring that into the old format

1day2die (changes requested):

resources/views/admin/products/index.blade.php
IceToast: ´done

PR #321, opened 2021-12-02
HOTFIX: Error when calculating price+tax has 3 decimals

1day2die:
There was an error with the pricing and paypal

When the Product costs 2.50€  and you have 5% tax, that would calculate to 0.125€ making the total 2.625€

paypal can obivously not work with 3 decimals behind the price.

Rounding this up fixed the error.


IceToast:
Needs immediate merge >.>

PR #327, opened 2021-12-13
Localization and multilanguage Support (German included)

1day2die:
TO TEST THIS PR YOU NEED TO PROVIDE THE `LOCALE=de` OPTION IN YOUR .env

The whole Dashboard now uses LocalizationStrings instead of normal Hardcoded Strings

Users can add their own translations by adding a new language .json file in /resources/lang

Ive done this in an 6h session without breaks. At some point even my German-skills suffered so please dont slap me when i made gramatical mistakes

Some things like the Content of the Notifications cannot be translated. This would need a whole rework of the notification-system


IceToast:
hot

PR #339, opened 2021-12-22
servers card UI redesign

IceToast:
This PR is a redesign of the server cards from the servers/index.

Screenshot:
image


1day2die:
Nice PR.

Please add new Translationstring to the en.json file

anondev-sudo:
Can you add also node to Location in `()`

IceToast:
> Can you add also node to Location in `()`? 

We discussed this suggestion and came to the conclusion that having the user to know the exact nodes name is unnecessary and could lead to confusion.

anondev-sudo:
> > Can you add also node to Location in `()`?
> 
> We discussed this suggestion and came to the conclusion that having the user to know the exact nodes name is unnecessary and could lead to confusion.

When users can choose node when making server, they should be able to know on which node is their server located. So when host owner announcements that maintenance will be on Node x, users don't know if they have the server there. And I don't see problem in adding it.

1day2die:
> > > Can you add also node to Location in `()`?
> > 
> > 
> > We discussed this suggestion and came to the conclusion that having the user to know the exact nodes name is unnecessary and could lead to confusion.
> 
> When users can choose node when making server, they should be able to know on which node is their server located. So when host owner announcements that maintenance will be on Node x, users don't know if they have the server there. And I don't see problem in adding it.

It would look very cluddered and is not worth the information.
I could only imagine adding it in the Hover pop-up @IceToast  ?
Other than that the Client usually only needs to know WHERE the node is located which will be selected on creation.

anondev-sudo:
> I could only imagine adding it in the Hover pop-up @IceToast ?

Good

IceToast:
> I could only imagine adding it in the Hover pop-up @IceToast ?

Thats a good idea. like another info icon on that card right next to the location?

1day2die:
> > I could only imagine adding it in the Hover pop-up @IceToast ?
> 
> Thats a good idea. like another info icon on that card right next to the location?

i wouldve placed it on the same popup as the specifications. But yeah, another icon could also work. you do you

IceToast:
image

PR #347, opened 2021-12-26
Throw error on registration if account exists on pterodactyl

1day2die:
image


anondev-sudo:
Can you throw error also when you try to link Discord account that is already linked?

1day2die:
thats a whole different story and something to create a new issue for^^

PR #348, opened 2021-12-27
Update user on cpgg updates user on pterodactyl

1day2die:
This will enable Password, Name, and email updating and syncing to pterodactyl from cpgg


anondev-sudo:
Maybe this should be optional and enabled by default

1day2die:
maybe in the future. we will see how people will react to this. If the users want tidfferent passwords on these platforms, they can change their Password on cpgg first and then on ptero.

PR #354, opened 2021-12-28
removed PAYPAL_EMAIL env & Info row in checkout

### LINKED ISSUE #353: Remove PAYPAL_EMAIL from env
# 🚀 Feature Request

### Is your proposal related to a problem?

*Provide a clear and concise description of what the problem is.
For example, "I can't change x and it makes me have to do manual work"*

I have noticed that PAYPAL_EMAIL in .env is useless. Its not passed to Paypal.

### Describe the solution you'd like

*Provide a clear and concise description of what you want to happen.
For Example, "Automate the changing of x so I don't have to do it manually"*

https://github.com/ControlPanel-gg/dashboard/blob/development/resources/views/store/checkout.blade.php#L52
here replace the email on checkout page to `$InvoiceSettings->company_mail` (email that is set in invoice settings).

The data on checkout page should be the same as the data that are used for PDF invoice. If they are not, it may make users feel confused 😕.

Also instead of MAIL_FROM_NAME, company name should be used.

### Additional context

*Is there anything else you can add about the proposal?
You might want to link to related issues here, if you haven't already.*

Discussed with 1Day2Die

IceToast:
Title.

Email is not being used besides the checkout page. Info row at Checkout page does not show any important information to the customer:

This PR closes #353.

Screenshot (after refactor):
image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,70 +1,76 @@
+### --- App Settings --- ###
 APP_NAME=Controlpanel.gg
 APP_ENV=production
 APP_KEY=
 APP_DEBUG=false
 APP_URL=http://localhost
-#list with timezones https://www.php.net/manual/en/timezones.php
+# List with timezones https://www.php.net/manual/en/timezones.php
 APP_TIMEZONE=UTC
+### --- App Settings End --- ###
 
+### --- Localization settings --- ###
 # If set to true, Language is chosen automatically depending on the users browserlanguage.
 DYNAMIC_LOCALE = false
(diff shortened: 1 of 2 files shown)


1day2die:
ew it looks so naked :(

anondev-sudo:
This does not look good

anondev-sudo:
I think that the details that were there before should be there, just use email that you set as invoice company email. Display there data which from invoice company settings

PR #408, opened 2022-02-22
Fix users table sorting

anondev-sudo:
Sort users by default by last_seen in the table


anondev-sudo:
@1day2die or @IceToast Can you reopen it?

PR #432, opened 2022-06-02
Added read all Button for Notifications

1day2die:


RTK23-dev:
Preety useless imo

1day2die:
ah its really annoying when you have like 10 notifications and have to click though all of them.
Also there was a Issue/Request open for that so why not.

PR #476, opened 2022-07-11
Fixed <form> and better code look

miniSpoXster777:
Please don't hit me for these PRs


1day2die:
actually im pretty sure that the password changing is part of the form.

Its just that the submit button is placed on the left and the change password on the right.

The way you changed the code, changing password would most likely not work anymore :)

miniSpoXster777:
Yeah i see now. It confused me opening <form> but closing it after a lot </div>. Thats why my IDE is shouting that form is not closed in the right place

Issue #392, opened 2022-02-08
Description does not work with line-breaks


partridge420x:
**Describe the bug 🐛**
When editing the description, it does not listen to line breaks

**To Reproduce**
1. Create a package
2. define the description using line-breaks

**Expected behavior**

When using linebreaks
it should
look like
this (as an example

**Screenshots**
![image](https://user-images.githubusercontent.com/26298829/152895144-f520d118-31c8-4000-9893-f93a2a5858d1.png)
![image](https://user-images.githubusercontent.com/26298829/152895419-c71052ed-a019-4206-ab2c-4160193a7869.png)

**Desktop (please complete the following information):**
 - OS: windows 10
 - Browser Opera GX
 - Version unknown (assumed to be latest)

**Additional context**
Having functioning line breaks would make things a lot easier to read for the customer

PR #485, opened 2022-07-18
Fixes line-breaks on Description when creation a server

AGuyNamedJens:
This fixes issue #392 and adds a simple style (AlpineJS styling) see [here](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#pre)

Please, if you know how to insert this as a CSS class. Tell me. Anyways have fun!


AGuyNamedJens:
Do note, that the longer descriptions are, the longer the less-longer descriptions will be to be UI friendly, so please, don't make one description 5 lines long and the other 2.. Unless all of your clients would be using phones only since it's mobile friendly as well.

1day2die (approved):
Tested

PR #501, opened 2022-08-01
Ticket System

SahrulGnwn:
image
image
image
image
image

### TICKET SYSTEM
_Client Can Create Support Ticket To Contact Admin/Mod_

**Feature:**

- Only user with role Admin or Mod can show all ticket list
- Notification Integration ( User & All Admin/Mod get simple notification when ticket create/reply )
- Mail Integration ( User & All Admin/Mod get mail when ticket create/reply )


sybetjeh:
godamn

PR #572, opened 2022-08-20
Delete server on 404 instead of giving error 500 to the user

ok236449:
This commit prevents erroring out on a user if it got deleted on pterodactyl only. The deletion is only triggered upon loading the "/servers" route as I did not want to cause possible data loss due to some error.

I didn´t do anything about the ActivityLog, if the server gets deleted this way, a normal log message is being saved (User X deleted a server Y). Ideally, I would want It to say something like "Server X deleted by the system - server not found on pterodactyl".

Please, use with caution and review this code well.

### DIFF
--- a/app/Classes/Pterodactyl.php
+++ b/app/Classes/Pterodactyl.php
@@ -287,7 +287,7 @@ class Pterodactyl
      * @param int $pterodactylId
      * @return mixed
      */
-    public static function getServerAttributes(int $pterodactylId)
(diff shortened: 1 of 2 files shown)


sybetjeh:
I'm not able to make this work.

I deleted the server on pterodactyl's side, however when I go to /servers I still get a 500.

ok236449:
> I'm not able to make this work.
> 
> I deleted the server on pterodactyl's side, however when I go to /servers I still get a 500.

What is the exact error? It worked fine on my end.

Issue #299, opened 2021-11-14
[feature] Block useractions via configuration settings


BolverBlitz:
# 🚀 Feature Request

### Is your proposal related to a problem?
No

*Provide a clear and concise description of what the problem is.
For example, "I can't change x and it makes me have to do manual work"*

There is currently no good solution to prevent abuse, mentioned in #298 that dosn´t need manual work.

### Describe the solution you'd like

*Provide a clear and concise description of what you want to happen.
For Example, "Automate the changing of x so I don't have to do it manually"*

There could be more settings in configuration like:
Prevent_New_Users - Disables the regestration
Prevent_New_Servers - Disables the generation of new servers
Prevent_Login - Disables login

### Additional context

*Is there anything else you can add about the proposal?
You might want to link to related issues here, if you haven't already.*

If those settings are also avaible via the API, the admin can use already existing or self written programs that could stop such abuse.
If there are 200% more users that usual, a program could disable registration for some time or if someone spams a lot of servers or the nodes experience peak load a program could stop users from adding new servers.

IceToast:
Done in #595

PR #595, opened 2022-11-03
New Options

Ferks-FK:
This PR adds some of the options suggested in #299.

## Todo
- [x] Block the registration of new users.
- [x] Block the creation of new servers.
- [ ] Disable Login.

Any idea of another option? :)

### DIFF
--- a/app/Http/Controllers/Api/UserController.php
+++ b/app/Http/Controllers/Api/UserController.php
@@ -251,6 +251,13 @@ class UserController extends Controller
             'email' => ['required', 'string', 'email', 'max:64', 'unique:users'],
             'password' => ['required', 'string', 'min:8', 'max:191'],
         ]);
+        
+        // Prevent the creation of new users via API if this is enabled.
+        if (!config('SETTINGS::SYSTEM:CREATION_OF_NEW_USERS', 'true')) {
+            throw ValidationException::withMessages([
+                'error' => "The creation of new users has been blocked by the system administrator."
+            ]);
+        } 
 
         $user = User::create([
             'name' => $request->input('name'),
--- a/database/seeders/Seeds/SettingsSeeder.php
+++ b/database/seeders/Seeds/SettingsSeeder.php
@@ -522,5 +522,19 @@ class SettingsSeeder extends Seeder
             'type'  => 'string',
             'description'  => 'Enables the updgrade/downgrade feature for servers'
         ]);
+        Settings::firstOrCreate([
+            'key'   => 'SETTINGS::SYSTEM:DISABLE_NEW_SERVERS',
+        ], [
+            'value' => "",
+            'type'  => 'boolean',
+            'description'  => 'Enable/disable the creation of new servers'
+        ]);
+        Settings::firstOrCreate([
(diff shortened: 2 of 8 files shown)


sybetjeh:
with the disable login, also log everyone out if you enable that (otherwie what would be the point?) and maybe add a whitelist

Ferks-FK:
> with the disable login, also log everyone out if you enable that (otherwie what would be the point?) and maybe add a whitelist

I also see no real reason to disable the login.

sybetjeh:
Maintenance? Panel for only yourself? Depends what you're gonna use it for

Ferks-FK:
> 

What kind of maintenance can users not login?
There must also be artisan maintenance.

app/Http/Controllers/Api/UserController.php
```diff
@@ -251,6 +251,13 @@ public function store(Request $request)
             'email' => ['required', 'string', 'email', 'max:64', 'unique:users'],
             'password' => ['required', 'string', 'min:8', 'max:191'],
         ]);
+
```
Ferks-FK, line 254: Make it impossible to create users via the API if the option is disabled on the front-end.

database/seeders/Seeds/SettingsSeeder.php
```diff
@@ -518,21 +518,21 @@ public function run()
         Settings::firstOrCreate([
             'key'   => 'SETTINGS::SYSTEM:ENABLE_UPGRADE',
         ], [
-            'value' =>"",
-            'type'  => 'string',
+            'value' =>  "false",
```
Ferks-FK, line 521: Change the type of a key in the db, and default values for the new options.

app/Http/Controllers/Api/UserController.php
```diff
@@ -251,6 +251,13 @@ public function store(Request $request)
             'email' => ['required', 'string', 'email', 'max:64', 'unique:users'],
             'password' => ['required', 'string', 'min:8', 'max:191'],
         ]);
+        
+        // Prevent the creation of new users via API if this is enabled.
+        if (!config('SETTINGS::SYSTEM:CREATION_OF_NEW_USERS', 'true')) {
+            throw ValidationException::withMessages([
+                'error' => "The creation of new users has been blocked by the system administrator."
+            ]);
+        }
```
Ferks-FK, line 260: Make it impossible to create users via the API if the option is disabled on the front-end.

database/seeders/Seeds/SettingsSeeder.php
```diff
@@ -518,21 +518,21 @@ public function run()
...
         ]);
         Settings::firstOrCreate([
-            'key'   => 'SETTINGS::SYSTEM:DISABLE_NEW_USERS',
+            'key'   => 'SETTINGS::SYSTEM:CREATION_OF_NEW_USERS',
         ], [
-            'value' => "",
+            'value' => "true",
             'type'  => 'boolean',
             'description'  => 'Enable/disable the creation of new users'
         ]);
```
Ferks-FK, line 538: Change the type of a key in the db, and default values for the new options.

PR #611, opened 2023-01-03
change nest description from string to text

einsz:
Fixes error 500 on long nest description


IceToast:
Looks good to me. However you need to change the migration down function to change the type back to its original.

database/migrations/2023_01_03_185102_update_make_description_text_in_nests_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('nests', function (Blueprint $table) {
+            $table->text('description')->change();
```
IceToast, line 29: Looks good to me. However you need to change the migration down function to change the type back to its original.

einsz: ah yes, i see. should now be all right

PR #635, opened 2023-01-17
Fix language file for partner values

GeckoBoy84:


1day2die:
Are the translateable strings in the blade files changed aswell?

GeckoBoy84:
They were changed in blade but not translation strings so this is to fix that

PR #648, opened 2023-01-22
[Fix] User creation settings

Ferks-FK:
A simple fix for the "creation of new users" checkbox, where it was unchecked on first startup of the panel.


IceToast (changes requested):
- Please rename every strings that mention "enable/disable" or "enable-disable" to either "enabled" or "disabled".
- Please rename the setting at the setting frontend page to start with the appropriate verb 

As we discussed on Discord it is not clear what is meant when setting "enable/disable..." to true.
This also means you can remove the info-box at the settings page as it should be clear then.

Ferks-FK:
Fix the change of descriptions.
And additionally removed the `laravel-mix` package, because the new theme addition now uses `vite`.

Issue #734, opened 2023-02-10
[Bug]: Tickets when closed


GeckoBoy84:
### What is your request about?

When a ticket is closed on the dashboard, the close button is still visible on the ticket list
![image](https://user-images.githubusercontent.com/67899387/218150552-02b62ece-ae31-4516-beae-9a152ed77120.png)

### Branch

main

### Solution idea

The button should be replaced by a reopen button, this button should also be present on the ticket page.

IceToast:
Fixed in #736

Issue #724, opened 2023-02-06
[Bug]: Store Page Name


GeckoBoy84:
### What is your request about?

Because controlpanel used to only be able to sell credits, the store header was the credit display name. However we now offer Server Slots which means it should be app name + store or similar.

### Branch

main

### Solution idea

"app name" + store or similar

### Additional Info

![image](https://user-images.githubusercontent.com/67899387/217064064-d986274a-4992-47f3-ad4c-bae3c88cfef3.png)

LogischJo:
Looks reasonable. I will look into this tomorrow. What about making this App name + store or optionally something made up in the settings.

GeckoBoy84:
My opinion would be app name + store

IceToast:
Fixed in #736

PR #736, opened 2023-02-11
fix: 🚑️ Wrong Store Page Name and add option to Reopen Tickets 

LogischJo:


LogischJo:
Fixes #724

LogischJo:
Newest commit also fixes #734

app/Http/Controllers/Moderation/TicketsController.php
```diff
@@ -33,9 +33,14 @@ public function show($ticket_id)
         return view('moderator.ticket.show', compact('ticket', 'ticketcategory', 'ticketcomments', 'server'));
     }
 
-    public function close($ticket_id)
+    public function changeStatus($ticket_id)
     {
         $ticket = Ticket::where('ticket_id', $ticket_id)->firstOrFail();
```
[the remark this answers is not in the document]
IceToast: I would love to see the described changed also done for the methods "delete" and "reply" but thats up to you and no need for this fix.

app/Http/Controllers/Moderation/TicketsController.php
```diff
@@ -145,7 +175,13 @@ public function blacklist()
 
     public function blacklistAdd(Request $request)
     {
+        try {
         $user = User::where('id', $request->user_id)->first();
```
IceToast, line 179: When using first(), it wont throw an exception when there is no user. It will just leave the variable empty. Should be changed to firstOrFail as its mandatory i guess?

app/Http/Controllers/Moderation/TicketsController.php
```diff
@@ -145,7 +175,13 @@ public function blacklist()
...
+        }
+        catch (Exception $e){
+            return redirect()->back()->with('warning', __('User not found on the server. Check the admin database or try again later.'));
+
+        }
         $check = TicketBlacklist::where('user_id', $user->id)->first();
```
IceToast, line 185: Same for here then, you could put both queries inside the try block

Issue #773, opened 2023-03-27
[Bug]: Cannot set settings value when current value is NULL


IceToast:
### What is your request about?

When the current value of a setting is NULL it cannot be set using the settings page in the frontend. The SettingsController parses the value type from the current value which breaks when it's NULL.

### Branch

development

PR #776, opened 2023-03-28
[Fix & Updates] Empty settings not saving | Dependency updates | Dockerfile update

AGuyNamedJens:
This PR Fixes Issue #773 where `gettype` was checking the wrong value, resulting in settings not being saved at all.

Also included package updates to be up-to-date.
Also includes small changes to make the Dockerfile up-to-date as well.


app/Settings/PterodactylSettings.php
```diff
@@ -6,11 +6,11 @@
 
 class PterodactylSettings extends Settings
 {
-    public string $admin_token;
+    public ?string $admin_token;
```
IceToast, line 9: Should the PterodactylSettings be nullable really? Without them controlpanel cannot work and will throw errors whenever accessing any page that depend on pterodactyl (like the overview, dashboard, server creation).

composer.json
IceToast, line 1: Are those updates of the dependencies testes, that they do not introduce any breaking changes? 
A fix for a bug should not really introduce a whole dependency upgrade imo.

docker/docker-compose.yml
IceToast, line 1: Why does this fix change the docker compose file?

app/Settings/PterodactylSettings.php
AGuyNamedJens: I guess I shouldn't have committed that either then, i'll remove it.

app/Settings/PterodactylSettings.php
```diff
@@ -44,9 +44,9 @@ public function getUrl(): string
     public static function getValidations()
     {
         return [
-            'panel_url' => 'required|string|url',
-            'admin_token' => 'required|string',
-            'user_token' => 'required|string',
+            'panel_url' => 'nullable|string|url',
```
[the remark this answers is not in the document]
AGuyNamedJens: I guess I shouldn't have committed that either then, i'll remove it. Probably thought it was easier for testing purposes yesterday..

composer.json
AGuyNamedJens: The updates aren't related to the bug fixes, but I've had no issues yet with the updates. Only on my installation using the docker composer where i also had to change app.php, but i left that out from this PR.

docker/docker-compose.yml
AGuyNamedJens: I don't know how to split PRs into multiple, shouldn't be too annoying as probably no one except me used the docker setup anyways..

app/Http/Controllers/Admin/SettingsController.php
```diff
@@ -109,7 +109,7 @@ public function update(Request $request)
         $settingsClass = new $settings_class();
 
         foreach ($settingsClass->toArray() as $key => $value) {
-            switch (gettype($value)) {
+            switch (gettype($request->input($key))) {
```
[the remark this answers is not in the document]
AGuyNamedJens: > could result in unwanted errors due to type mismatches
If i change the switch statement's default type to just throw in the value that was given instead of setting it to null, as well as adding 'NULL' as it's own switch case, then your comment would be invalidated, and we would shift the "unwanted errors" to not be an issue on our side. Effectively moving the blame to "Code your value types in the front end better" towards whoever changes the front-end, as the backend will not reject other types with the change I could propose..

PR #778, opened 2023-03-29
(Fix/Feat/Refactor) Installer|Functions|Database|Settings

AGuyNamedJens:
This PR Modifies 14 files; by refactoring, fixing bugs and adding new features:

 **Installer.php** for first-time setups:
- Fixed API Calls
- Fixed Database modification requests
- Fixed encryption to match the encryption used by the database
- Fixed deprecation warnings, changed several checks and functions
- Added debug logs
------
**Functions.php:**
- Adds encryption and decryption functions
- Refactor and Typedoc everything
- Modified environment functions to work properly
- New, renamed, refactored and removed functions
- Added debug logs
------
**Index.php:**
- Refactor, visual improvements to the codebase. No actual code changed.
------
**Settings controller:**
- Refactored validation rules to match input options
- Refactored variables to match validation rules
------
Removed the `main_site` from the database since it's no longer used

### DIFF
--- a/public/install/functions.php
+++ b/public/install/functions.php
@@ -31,7 +31,7 @@ function getMySQLVersion()
 {
     global $requirements;
 
-    $output = shell_exec('mysql -V');
+    $output = shell_exec('mysql -V') ?? '';
     preg_match('@[0-9]+\.[0-9]+\.[0-9]+@', $output, $version);
 
     $versionoutput = $version[0] ?? '0';
@@ -41,7 +41,7 @@ function getMySQLVersion()
 
 function getZipVersion()
 {
-    $output = shell_exec('zip  -v');
+    $output = shell_exec('zip  -v') ?? '';
     preg_match('@[0-9]+\.[0-9]+\.[0-9]+@', $output, $version);
 
     $versionoutput = $version[0] ?? 0;
@@ -51,7 +51,7 @@ function getZipVersion()
 
 function getGitVersion()
 {
-    $output = shell_exec('git  --version');
+    $output = shell_exec('git  --version') ?? '';
     preg_match('@[0-9]+\.[0-9]+\.[0-9]+@', $output, $version);
 
     $versionoutput = $version[0] ?? 0;
@@ -61,7 +61,7 @@ function getGitVersion()
 
 function getTarVersion()
 {
-    $output = shell_exec('tar  --version');
+    $output = shell_exec('tar  --version') ?? '';
     preg_match('@[0-9]+\.[0-9]+@', $output, $version);
 
     $versionoutput = $version[0] ?? 0;
--- a/public/install/functions.php
+++ b/public/install/functions.php
@@ -266,7 +266,6 @@ function wh_log(string $message, string $level = 'info', array $context = []): v
 
     switch (strtolower($level)) {
         case 'debug': // Only log debug messages if APP_DEBUG is true
-            wh_log('APP_DEBUG: ' . getEnvironmentValue('APP_DEBUG'));
             if(getEnvironmentValue('APP_DEBUG') === false) return;
             $log->debug($message, $context);
             break;
--- a/public/install/functions.php
+++ b/public/install/functions.php
@@ -3,7 +3,9 @@ require '../../vendor/autoload.php';
 
 use Illuminate\Encryption\Encrypter;
 use Illuminate\Support\Str;
-use Illuminate\Support\Facades\Log;
+use Monolog\Formatter\LineFormatter;
+use Monolog\Handler\StreamHandler;
+use Monolog\Logger;
 
 $required_extensions = ['openssl', 'gd', 'mysql', 'PDO', 'mbstring', 'tokenizer', 'bcmath', 'xml', 'curl', 'zip', 'intl'];
 
@@ -20,6 +22,8 @@ $requirements = [
 function checkPhpVersion(): string
 {
     global $requirements;
+
+    wh_log('php version: ' . phpversion(), 'debug');
     if (version_compare(phpversion(), $requirements['minPhp'], '>=') && version_compare(phpversion(), $requirements['maxPhp'], '<=')) {
         return 'OK';
     }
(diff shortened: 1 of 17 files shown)


public/install/functions.php
```diff
@@ -17,126 +25,196 @@ function checkPhpVersion()
...
 }
 
-function setEnvironmentValue($envKey, $envValue)
+/**
+ * Sets the environment variable into the env file
+ * @param string $envKey The environment key to set or modify
+ * @param string $envValue The environment variable to set
+ * @return bool true on success or false on failure.
+ */
+function setEnvironmentValue(string $envKey, $envValue)
```
IceToast, line 132: Just using putenv/getenv and put it into a new function seems utterly redundant to me.

public/install/functions.php
```diff
@@ -17,126 +25,196 @@ function checkPhpVersion()
...
+/**
+ * Gets the variable from the env file
+ * @param string $envKey The environment variable to look for
+ * @return array|false|string Returns the value if found, otherwise returns false.
+ */
+function getEnvironmentValue(string $envKey): array|false|string
 {
-    $envFile = dirname(__FILE__, 3).'/.env';
-    $str = file_get_contents($envFile);
+    return getenv($envKey);
```
IceToast, line 145: See above

public/install/functions.php
```diff
@@ -17,126 +25,196 @@ function checkPhpVersion()
...
+ * Encrypt the given value
+ * @param mixed $value The variable to be encrypted
+ * @param bool $serialize If the encryption should be serialized
+ * @return string Returns the encrypted variable.
+ */
+function encryptSettingsValue(mixed $value, $serialize = true): string
+{
+    $appKey = getEnvironmentValue('APP_KEY');
+    $appKey = base64_decode(Str::after($appKey, 'base64:'));
+    $encrypter = new Encrypter($appKey, 'AES-256-CBC');
```
IceToast, line 159: The cipher may also not be provided since its the default anyways

public/install/functions.php
```diff
@@ -17,126 +25,196 @@ function checkPhpVersion()
...
+ * @param mixed $payload The payload to be decrypted
+ * @param bool $unserialize If the encryption should be unserialized
+ * @return mixed Returns the decrypted variable on success, throws otherwise.
+ */
+
+function decryptSettingsValue(mixed $payload, $unserialize = true)
+{
+    $appKey = getEnvironmentValue('APP_KEY');
+    $appKey = base64_decode(Str::after($appKey, 'base64:'));
+    $encrypter = new Encrypter($appKey, 'AES-256-CBC');
```
IceToast, line 176: See above

public/install/functions.php
```diff
@@ -17,126 +25,196 @@ function checkPhpVersion()
...
+    return $decryptedKey;
 }
 
-function run_console($command)
+/**
+ * Run a shell command
+ * @param string $command The command string to run
+ * @return false|string|null Returns the result from the command.
+ */
+function run_console(string $command)
```
IceToast, line 187: This should be refactored. It does only support the bash shell which won't be present on windows or some specific linux distros. Rather use php shell_exec for example:
https://www.php.net/manual/en/function.shell-exec.php

public/install/functions.php
```diff
@@ -17,126 +25,196 @@ function checkPhpVersion()
...
+function wh_log(string $log_msg)
 {
     $log_filename = 'logs';
-    if (! file_exists($log_filename)) {
+    if (!file_exists($log_filename)) {
         // create directory/folder uploads.
         mkdir($log_filename, 0777, true);
     }
-    $log_file_data = $log_filename.'/installer.log';
+    $log_file_data = $log_filename . '/installer.log';
```
IceToast, line 207: When i think about, we should put the installer log at the same place where the laravel log lives in. (storage/logs)

IceToast (changes requested):

public/install/functions.php
AGuyNamedJens: That was there because i had not refactored the actual functions in each file yet

public/install/functions.php
AGuyNamedJens: I agree.

public/install/functions.php
AGuyNamedJens: I've refactored getenv, removed the function. Refactored setenv however, due to the way the code looks using "=" is not ideal. I prefer to keep the function method for setting environmental values.

public/install/functions.php
AGuyNamedJens: Removed.

public/install/functions.php
AGuyNamedJens: Are we really going to make sure people can host this dashboard on windows when even pterodactyl themselves do not offer support for windows hosting?

1day2die: no. we do not offer windows support. its never mentioned anywhere. we do somewhere say it works on ubuntu, debian and centos. everything else is not our business

AGuyNamedJens (commented):
I haven't touched the general code, but this function seems to be working much better and has more options such as work directory, the outputs to listen to, customizability options which change behavior on windows and more. Refer to https://www.php.net/manual/en/function.proc-open.php proc_open for more details on the features.

AGuyNamedJens (commented):
The new logging system needs a test to see if it works, at the time of the commit I was unable to test this.

IceToast:
LGTM, please check if the logging works. Can be merged if it does

public/install/functions.php
```diff
@@ -102,41 +169,114 @@ function setEnvironmentValue($envKey, $envValue)
...
+ * @return void
+ */
+function wh_log(string $message, string $level = 'info', array $context = []): void
 {
-    $log_filename = 'logs';
-    if (! file_exists($log_filename)) {
-        // create directory/folder uploads.
-        mkdir($log_filename, 0777, true);
+    $formatter = new LineFormatter(null, null, true, true);
+    $stream = new StreamHandler(dirname(__FILE__, 3) . '/storage/logs/controlpanel.log', Logger::DEBUG);
```
1day2die, line 246: change to installer.log

AGuyNamedJens: Oh alright, will do that.

PR #803, opened 2023-05-02
Mobile Rotate Bug

2IceCube:
Bug Fix for old IOS Devices Rotate Bug


IceToast:
What exactly has been changed in this PR to achive that fix?

AGuyNamedJens:
@2IceCube Could you explain in more detail what has changed and why (what "old IOS Device" for example?)

PR #854, opened 2023-05-30
[Fix] Replace old demo link

AGuyNamedJens:
The demo URL still links to the old domain, thus not working. This commit fixes this by providing the correct domain.


1day2die:
already done in dev branch

AGuyNamedJens:
> already done in dev branch

Yes but the dev branch isn't the one people use, it was a support question in the discord so i changed it here..
We should fix such issues on the main branch instead of dev only

PR #867, opened 2023-06-08
Correction of incorrect text output

MrWeez:


MrWeez:
Before the change
image

After the change
image

PR #943, opened 2024-04-12
Update app.php

MrWeez:
Update app.php to make users see a message about the need to upgrade to a new version, to fix an error with alpinejs

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -4,7 +4,7 @@ use Illuminate\Support\Facades\Facade;
 
 return [
 
-    'version' => '0.9.6',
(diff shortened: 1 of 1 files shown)


1day2die:
has anything changed?

MrWeez:
> has anything changed?

Ferks updated the link to alpinejs as [old](https://unpkg.com/alpinejs@3.13.8/dist/cdn.min.js%E2%80%9D ) stopped working. Therefore, it seems to me it is worth notifying users about the new version in order to avoid a bunch of support requests.

PR #951, opened 2024-05-14
fix: Panel Installer

S0ly:
I invite contributions to address several issues with the installer for the ctrlpanel:

### Issues to be Fixed:

- [x] **Unconfigured Panel Handling**: Currently, the ctrlpanel displays an error when initiated without an install.lock file. It could be enhanced by implementing a mechanism to detect the absence of install.lock and redirecting users to /install to initiate the installation process.

- [x] **Deprecated PHP Function Error in trustedproxy.php**: The consistent appearance of the error message "Deprecated: explode(): Passing null to parameter ($string) of type string is deprecated in /var/www/html/config/trustedproxy.php on line 29" necessitates immediate resolution to prevent deprecated function calls.

- [x] **Docker MySQL and Unnecessary Checks**: The Docker setup lacks MySQL, which may be unnecessary for some users as they store their databases outside the server. Additionally, HTTPS checks are redundant as modern web browsers already provide warnings.

- [x] **Failed PHP Extension Check for Redis**: Despite the potential necessity of Redis, the PHP extension check fails to verify its presence. This issue requires rectification to ensure proper functionality.

- [x] **Inconsistent Behavior with php artisan migrate --seed --force**: Occasional failures of this command, accompanied by deprecated errors and timezone-related notices, demand investigation to ensure seamless execution.

- [x] **Post-Installation Issues with Redis Configuration**: Following installation, the ctrlpanel encounters difficulties connecting to Redis, necessitating manual configuration in the .env file.

- [x] **Unencrypted Tokens in Settings Table**: The installer sets tokens directly in the settings table without encryption, causing errors when Laravel attempts to decrypt them. This requires modification to ensure tokens are encrypted properly during installation.

Contributors are encouraged to address these issues to enhance the reliability and usability of the ctrlpanel installer. Thank you for your contributions!


jameskitt616:
I'd like to take a look and implement "Post-Installation Issues with Redis Configuration" I assume it will be an extra step after/before database configuration and preferably with a verification that the redis server connection was successful?

S0ly:
@jameskitt616 yep exactly

S0ly:
I think the best would be after the main database its just some env value to set and test

IceToast:
LGTM, couldnt test it for sure. Please fix conflicts too.

kenshin133:
idk how to comment on specific fixes but checking this out I found a couple things:
Unconfigured Panel Handling: this seems completely fixed, tried a few ways and could not get un-favorable results.

during install I still got this `  at vendor/laravel/framework/src/Illuminate/Encryption/EncryptionServiceProvider.php:79
     75▕     protected function key(array $config)
     76▕     {
     77▕         return tap($config['key'], function ($key) {
     78▕             if (empty($key)) {
  ➜  79▕                 throw new MissingAppKeyException;
     80▕             }
     81▕         });
     82▕     }
     83▕ }

  i   Your app key is missing:
      https://laravel.com/docs/master/installation#configuration`
      and had to click install a second time.

IceToast:
@S0ly Should the installer handle the app key generation too?

S0ly:
> @S0ly Should the installer handle the app key generation too?

I suppose that if he can he should no ? like if no app key is present maybe generate one ? since he do all the rest, but if you think it should not be handled by the installer we could have the script of weez doing it, since we now have an installation script

jameskitt616:
Wasn't that fixed somewhere?
When i was re-factoring the web-installer i didn't ran into the issue anymore.

S0ly:
its why I called it " Inconsistent Behavior" sometimes it happend sometimes not

jameskitt616:
Not really, if it was a permission issue, it would need to fail everytime.

jameskitt616:
I (think) i fixed this issue with my timezone fix with the new Installer

jameskitt616:
I found out if there is a fresh install with no `APP_KEY`, the first time running the installer and generating the key fails for some reason, will look into it and fix it.

jameskitt616:
@S0ly please check the new PR with fixes https://github.com/S0ly/panel/pull/6

S0ly:
@1day2die @IceToast this PR should be done it still need some testing to be sure everything work as excepted

Issue #834, opened 2023-05-12
[Feature] Redirect to store page


GeckoBoy84:
### Feature Description

When the user does not have enough credits to purchase a plan, the button is unclickable.

### Solution Idea

Redirect them to the store page

S0ly:
have this been done ? I think I can do it :)

kenshin133:
in 0.9.8 : the behavior is interesting but not insane :
![image](https://github.com/Ctrlpanel-gg/panel/assets/29851415/d5f5812c-3fe8-48c5-97f9-e5398b6c9474)
one node is too big, one I can afford and one I cant. 

If I click the one I cant afford I get : 
![image](https://github.com/Ctrlpanel-gg/panel/assets/29851415/6e57b2a4-d379-45d5-9e32-c33ca0edd42b)
I would hope for 1. this to take me to the shop, or what might be more smooth, to have a second button that lets me take myself to the shop. 
![image](https://github.com/Ctrlpanel-gg/panel/assets/29851415/4a8c3cbe-2fdc-4eca-a7f7-a1a317e24498)


WIll update with 0.10.0 results soon

S0ly:
this is a medium priority so it will not be fixed in 0.10
but it may be fixed just after

S0ly:
except if someone want to do it since its labeled as a good first issue :) maybe you can do it ?

kenshin133:
in 0.10 the result is similar 
![image](https://github.com/Ctrlpanel-gg/panel/assets/29851415/de66fb63-f2d9-416e-aeaa-5c40ac708426)

I thing this is working as designed, but the change should be to make it more seamless to get people to buy more coins/tokens/credits. 

Personally I would not replace the "not enough credits" with a buy link, but perhaps a third block, or second button to leads to the store page.

MrWeez:
@S0ly can you assign me to this issue? I can implement this.

S0ly:
@MrWeez 👍

MrWeez:
> @MrWeez 👍

Tysm ❤️

MrWeez:
Added in the https://github.com/Ctrlpanel-gg/panel/pull/956

MrWeez:
@S0ly close this issue as completed

S0ly:
@MrWeez ho right, next time check on your PR you can add issues related that they close automaticly whit your PR if I remember well :)

S0ly:
[...]

PR #956, opened 2024-05-22
"Buy more" button, if user does not have enough credits

MrWeez:
Closes https://github.com/Ctrlpanel-gg/panel/issues/834


S0ly:
seems right !

PR #957, opened 2024-05-22
fixed product description showing up instead of name in two locations

kenshin133:
This seems to be referenced two places, admin/servers and admin/users/1

I could not find any other effects of this change but do feel free to double check my work.

themes/default/views/admin/servers/table.blade.php
And
themes/default/views/admin/users/show.blade.php
and perhaps
themes/default/views/admin/users/edit.blade.php

have
data: 'resources', name: 'product.name',

In them, another fix would be to alter this to data: 'product.name' but the proposed change seems like the right one.

### DIFF
--- a/app/Http/Controllers/Admin/ServerController.php
+++ b/app/Http/Controllers/Admin/ServerController.php
@@ -228,7 +228,7 @@ class ServerController extends Controller
                 return '<a href="' . route('admin.users.show', $server->user->id) . '">' . $server->user->name . '</a>';
             })
             ->addColumn('resources', function (Server $server) {
-                return $server->product->description;
+                return $server->product->name;
             })
             ->addColumn('actions', function (Server $server) {
                 $suspendColor = $server->isSuspended() ? 'btn-success' : 'btn-warning';


S0ly:
And don’t make commit that long 😂 add a comment here if needed

S0ly:
But other than that it’s fine :) thanks for the fix !

kenshin133:
Sorry, thought that was for the pull, I dont see a way to edit it, the code is ready. so either send it through or I can make another PR

S0ly:
Don’t worry you seems new :D and we don’t have good contributing guidelines for the moment 

For me the change seems fine but I would like someone else to check it since I’m confused by what need to be shown

MrWeez:
> Don’t worry you seems new :D and we don’t have good contributing guidelines for the moment
> 
> For me the change seems fine but I would like someone else to check it since I’m confused by what need to be shown

I'll check it out when I get home.

S0ly:
Thanks tell me if it’s good ;)

MrWeez:
> Thanks tell me if it’s good ;)

Yep, everything works. You can merge this PR

Issue #961, opened 2024-05-23
[Bug]: Error 500 if the price is more than 999 on the checkout


MrWeez:
### What is your request about?

If the price of the product in the store is more than 999, an error 500 occurs. I already have a solution to this problem, and I will do a PR soon

### Branch

main

### Solution idea

I'll fix it within an hour, probably

### Ctrlpanel Logs

```Shell
[previous exception] [object] (TypeError(code: 0): NumberFormatter::formatCurrency(): Argument #1 ($amount) must be of type float, string given at /var/www/controlpanel/app/Models/ShopProduct.php:59)
[stacktrace]
```

AGuyNamedJens:
Resolved with #963

PR #962, opened 2024-05-23
Fix 500 error if price more than 999

MrWeez:
This PR fixes and closes https://github.com/Ctrlpanel-gg/panel/issues/961

### DIFF
--- a/app/Models/ShopProduct.php
+++ b/app/Models/ShopProduct.php
@@ -62,8 +62,8 @@ class ShopProduct extends Model
     public function formatToCurrency($value, $locale = 'en_US')
     {
         $formatter = new NumberFormatter($locale, NumberFormatter::CURRENCY);
-
-        return $formatter->formatCurrency($value, $this->currency_code);
+    
+        return $formatter->formatCurrency((float) $value, $this->currency_code);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/Models/ShopProduct.php
```diff
@@ -62,8 +62,8 @@ public static function boot()
     public function formatToCurrency($value, $locale = 'en_US')
     {
         $formatter = new NumberFormatter($locale, NumberFormatter::CURRENCY);
-
-        return $formatter->formatCurrency($value, $this->currency_code);
+    
+        return $formatter->formatCurrency((float) $value, $this->currency_code);
```
IceToast, line 66: We cannot use float for currency calculations

S0ly:
@IceToast the fix seems correct based on the error in the issues why we cant use float ?

app/Models/ShopProduct.php
MrWeez: Hm, ok. But why? Could you please tell me in more detail, since the logs say that the float value is expected

```
[previous exception] [object] (TypeError(code: 0): NumberFormatter::formatCurrency(): Argument #1 ($amount) must be of type float, string given at /var/www/controlpanel/app/Models/ShopProduct.php:59)
[stacktrace]```

MrWeez:
@IceToast What about the latest commit? Will it do?

PR #952, opened 2024-05-17
fix: Panel Inconsistencies and issues

S0ly:
This branch focuses on resolving critical issues to prepare for production deployment. Below is the list of issues that need to be addressed:

- [x] csrf forms miss the hidden input and cause a 419 error
- [x] #933
- [x] #888 
- [x] #886 
- [x] #935 
- [x] #954 
- [x] #988
- [x] #989
- [x] #960 (fix roles) 
- [x] #958 
- [x] #971 
- [x] #970 
- [x] #1011
- [x] #1013

If you discover additional issues, please inform me, and I will add them to this list.

Additionally, the codebase has some inconsistencies, such as unclosed HTML tags. Contributors are encouraged to fix these minor bugs and make any improvements where possible.

Your efforts to enhance the reliability and usability of the ctrlpanel are greatly appreciated!

Thank you for your contributions!

PR #976, opened 2024-06-22
Move transferusers

jameskitt616:
Move transferusers file for #952 also Updated docs see: https://github.com/Ctrlpanel-gg/docs/pull/117

### DIFF
--- a/app/Helpers/TransferUsers.php
+++ b/app/Helpers/TransferUsers.php
@@ -13,9 +13,9 @@ function generateRandomString($length = 8) {
     for ($i = 0; $i < $length; $i++) {
         $randomString .= $characters[rand(0, $charactersLength - 1)];
     }
+
     return $randomString;
 }
-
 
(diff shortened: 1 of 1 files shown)


IceToast:
It is not part of the App (which is the laravel app) imo. Actually the transfer users script should ultimately not exist and be a laravel command (app/kernel/commands)

jameskitt616:
So delete the script and only rely on this? https://market.ctrlpanel.gg/resources/resource/35-transfer-users-from-ptero-to-cpgg/

IceToast:
Both options are not ideal

jameskitt616:
I do agree, a laravel command would make the most sense. No clue how to write one tho.

jameskitt616:
So what happens next? Do i close the PR and trash it? & Remove the Docs PR.

S0ly:
@IceToast

IceToast:
I would not move it. I would keep it as is (a temporary solution) that gets replaced with a laravel command later on

PR #983, opened 2024-07-25
Fixed the problem with permissions and roles

Zastinian:
Fixed the problem caused by updating spatie dependencies.


Zastinian:
I just fixed the design and the error that the roles were not shown in the admin section.

1day2die:
LGTM. If ice approves it can be merged

app/Http/Controllers/Admin/RoleController.php
```diff
@@ -72,7 +72,8 @@ public function store(Request $request): RedirectResponse
         ]);
 
         if ($request->permissions) {
-            $role->givePermissionTo($request->permissions);
+            $collectedPermissions = collect($request->permissions)->map(fn($val)=>(int)$val);
```
IceToast, line 75: Why need the map items to be casted to an int?

Zastinian: https://spatie.be/docs/laravel-permission/v6/upgrading#content-upgrading-from-v5-to-v6

Screenshot_20240726_064658_com android chrome

1day2die: @IceToast

PR #987, opened 2024-07-29
Fixed the problem with ticket middleware

Zastinian:
ticket-new and ticket-reply are not defined in the providers so it causes problems in laravel 11.

image
image


MrWeez:
@IceToast can you please review this PR and merge it. Thanks

AGuyNamedJens (commented):
LGTM. Would recommend to make a middleware though

IceToast (approved):

Issue #1022, opened 2024-11-11
[Bug]: Timeout errors when sending emails, syncing Pterodactyl/servers


Huguitis:
### What is your request about?

When sending a lot of emails, the server takes so much to process the request and returns a timeout error. This also happens with Pterodactyl Sync and server sync when the API takes longer than expected.

### Branch

main

### Solution idea

Don't await this kind of tasks.

### Ctrlpanel Logs

```Shell
https://ptero.co/ihayofuqup
```

### Additional Info

![image](https://github.com/user-attachments/assets/f53df494-f9c8-47d6-9a1f-672315cf17b3)

lezetho:
Unfortunately this is common when there's a large amount of users that ctrlpanel has to email. What I would suggest the dev team to do is either increase the timeout or redirect them back, tell them it's sending, and send the emails in the background.

SuperEvilLuke:
As a temporary fix you could edit the following line to increase the timeout: https://github.com/Ctrlpanel-gg/panel/blob/4358d071b36ff5e9faf21cbe5e01289443163a98/public/install/phpmailer/SMTP.php#L164

MrWeez:
> As a temporary fix you could edit the following line to increase the timeout:
> 
> https://github.com/Ctrlpanel-gg/panel/blob/4358d071b36ff5e9faf21cbe5e01289443163a98/public/install/phpmailer/SMTP.php#L164

This file is responsible for sending emails only during web installation (/install page) and is not related to sending emails in Laravel

SuperEvilLuke:
ah darn, mb

PR #1080, opened 2024-12-21
1022

Ferks-FK:
Possible fix for #1022, must be tested first.


Ferks-FK:
Conflits with another PR.

Ferks-FK:
I think it's all right, opening it again.

PR #1083, opened 2024-12-22
fix: #1022

### LINKED ISSUE #1022: [Bug]: Timeout errors when sending emails, syncing Pterodactyl/servers

Ferks-FK:
Fixed #1022 (must be tested first)

### DIFF
--- a/app/Notifications/DynamicNotification.php
+++ b/app/Notifications/DynamicNotification.php
@@ -4,8 +4,9 @@ namespace App\Notifications;
 
 use Illuminate\Bus\Queueable;
 use Illuminate\Notifications\Notification;
+use Illuminate\Contracts\Queue\ShouldQueue;
 
-class DynamicNotification extends Notification
+class DynamicNotification extends Notification implements ShouldQueue
 {
     use Queueable;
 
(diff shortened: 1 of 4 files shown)


MrWeez:
Oh, I didn't even notice this PR. The only problem is that we need someone who has a lot of users in CtrlPanel. Well, or as another option, create fake users ¯\_(ツ)_/¯

S0ly (approved):

Ferks-FK:
> Oh, I didn't even notice this PR. The only problem is that we need someone who has a lot of users in CtrlPanel. Well, or as another option, create fake users 🤷

In fact, it would be nice to do tests with fake users.
This fix just throws the job of notifying onto the worker, which won't crash the application, and won't cause timeout errors.
Although I'm sure it solves most of these problems.

1day2die:
php artisan tinker

User::factory()->count(2000)->create();

PR #1084, opened 2024-12-24
fix: transferusers.php

RaphaelCapone:
---

💡 **Description**

Updating transferusers.php for the new database changes. Now tranfered users by this script are listed properly in the admin panel and can use services. see issue at: https://discord.com/channels/787829714483019826/850551820018384907/1320901027649683506

---

🛠️ **Type of Change**

Breaking change

---
* Change tested *


S0ly:
why you say it contain breaking changes ? your only modifying a third party script

S0ly:
also checking this script / code it contains many things wrong, in relation to the fix or not

MrWeez:
@RaphaelCapone Thanks for the PR, but could you please make these changes in the development branch instead of the main? Also, you added a php closing tag, which is not correct according to the norms of writing code in php.

MrWeez:
Otherwise, everything looks correct, good job

S0ly:
Correct $CPPPG_PASSWORD to $CPGG_PASSWORD.
also why the roles are selected by ID ? is it not like the worst things to do ???
I mean I understand that he is only fixing one problem but this script is full of problems

MrWeez:
> also why the roles are selected by ID ? is it not like the worst things to do ???

No, because the default role IDs are not changeable. They are the same for all instances

S0ly:
yes but you still should never use IDs 99% of the time. its just something that wait to break here

S0ly:
anyway @RaphaelCapone you can only do your fix, if you dont want to fully fix this code I may refactor it in the future, thanks you

MrWeez:
> yes but you still should never use IDs 99% of the time. its just something that wait to break here

In that case, how do you see the implementation of this? This is how the permission system works, and to fix it you need to change the backend of the CtrlPanel itself

S0ly:
if how the backend is made

but if you have a mysql row saying, role name = this string

you can search by name and not by ID, IDs are normally automatically generated by mysql and are unpredictable 

anyway

MrWeez (approved):
LGTM, but I'll wait for the review from @1day2die , he's more familiar with his script after all 😅

S0ly:
I think we should accept this fix and I will further refactor and fix this script, also this script have nothing to do here, it should be in the app/Console/Commands directory for laravel

RaphaelCapone:
> yes but you still should never use IDs 99% of the time. its just something that wait to break here

I know fields are kinda confusing, but the thing is that, in my change, after the new user is imported it just gets its ID(not setting it manually or anything) in order to set the right ID in the `model_id` field.
So, for example if ctrl panel have this user table(maybe the some users got deleted):
id name
1 user1
4 user2  ----> (the new user imported)

The model_has_roles table will look like that
id role_id model_type            model_id
1        1   App\Models\User      1
2        3  App\Models\User       4 --> the Localy ID of the imported ptero panel 

Hope my explanation will clarify things up

Issue #1059, opened 2024-11-26
Server creation is returning 'no available nodes' even when the first node is free but not permitted for the specified product.


BolverBlitz:
### What is your request about?

You have 2 nodes on the same location, but only node 2 is avaible for the product you wanna buy then you do not get any node.

on getAvaibleNode the arguments passed are location and product, the collection var that should hold only nodes the product is allowed on gets filled by a SQL query with only the where condition of location_id. 
The first avaible node gets returned by the function, so if the first node is avaible but the product is not allowed it would show that there are no avaible nodes.

### Branch

main

### Solution idea

I think a query like this should fix the issue
```
SELECT nodes.*
FROM nodes
JOIN node_product ON nodes.id = node_product.node_id
WHERE nodes.location_id = :location_id
  AND node_product.product_id = :product_id
ORDER BY nodes.id
```
If there isn´t any PHP magic already doing it, also limit the SELECT on id as thats the only value you use.
Also overwriting a function parameter, talking about checkNodeResources, is usualy considered poor coding practice, on that note when only the node->id is used anyway, why pass the entire node object?

1day2die:
fixed. thanks

PR #1096, opened 2025-01-06
Fix nodes

Ferks-FK:
This PR fixes two critical problems.
1. The server being allocated to a fully booked node, even when there was another node with resources available.
2. Corrects the verification of nodes that were not associated with the product.

It should also finally be fixed: #1059


AGuyNamedJens (approved):
Looks good to me

PR #1128, opened 2025-01-15
Recaptcha V3 Support || TOS Route Fix on Register Page || Dynamic Theme name loading in Settings

1day2die:
This PR adds support for Recaptcha V3, while also keeping V2 as posibility.

Also it fixes a route error @Ferks-FK  has overlooked kekw

Also this implements dynamic loading of themes to select on the settings page including validations

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -66,7 +66,7 @@ class LoginController extends Controller
             if($general_settings->recaptcha_version === "v2") {
                 $validationRules['g-recaptcha-response'] = ['required', 'recaptcha'];
             }elseif($general_settings->recaptcha_version === "v3") {
-                $validationRules['g-recaptcha-response'] = ['required','recaptchav3:recaptchathree,1.0'];
+                $validationRules['g-recaptcha-response'] = ['required','recaptchav3:recaptchathree,0.5'];
             }
         }
         $request->validate($validationRules);
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -63,10 +63,13 @@ class LoginController extends Controller
             'password' => 'required|string',
         ];
         if ($general_settings->recaptcha_version) {
-            if($general_settings->recaptcha_version === "v2") {
-                $validationRules['g-recaptcha-response'] = ['required', 'recaptcha'];
-            }elseif($general_settings->recaptcha_version === "v3") {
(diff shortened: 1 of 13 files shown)


app/Http/Controllers/Auth/LoginController.php
```diff
@@ -56,12 +56,18 @@ public function username()
...
     {
+
+
         $validationRules = [
             $this->username() => 'required|string',
             'password' => 'required|string',
         ];
-        if ($general_settings->recaptcha_enabled) {
-            $validationRules['g-recaptcha-response'] = ['required', 'recaptcha'];
+        if ($general_settings->recaptcha_version) {
```
IceToast, line 65: No need for the first if statement.
Would rather like a switch statement for other versions

app/Http/Controllers/Auth/RegisterController.php
```diff
@@ -96,8 +97,12 @@ protected function validator(array $data)
             'email' => ['required', 'string', 'email', 'max:64', 'unique:users'],
             'password' => ['required', 'string', 'min:8', 'confirmed'],
         ];
-        if ($this->recaptcha_enabled) {
-            $validationRules['g-recaptcha-response'] = ['required', 'recaptcha'];
+        if ($this->recaptcha_version) {
```
IceToast, line 100: Same same

themes/default/views/auth/login.blade.php
```diff
@@ -126,6 +131,7 @@ class="form-control @error('password') is-invalid @enderror"
                   | <a target="_blank"
                       href="{{ route('terms', 'tos') }}"><strong>{{ __('Terms of Service') }}</strong></a>
               @endif
+                {{Config::get('recaptchav3.api_site_key')}} ---- {{Config::get('recaptchav3.api_secret_key')}}
```
IceToast, line 134: Not sure if we should print out the site and secret key to the user lol

themes/default/views/layouts/app.blade.php
```diff
@@ -28,8 +30,13 @@
     <noscript>
         <link rel="stylesheet" href="{{ asset('plugins/fontawesome-free/css/all.min.css') }}">
     </noscript>
-    @if (app(App\Settings\GeneralSettings::class)->recaptcha_enabled)
+    @php ($recaptchaVersion = app(App\Settings\GeneralSettings::class)->recaptcha_version)
+    @if ($recaptchaVersion)
```
IceToast, line 34: Same if related thing as above

IceToast (changes requested):

PR #1157, opened 2025-02-08
Fix discord id searching in users table

DominikSLK:
💡 **Description**

Fixes/adds back searching using discord id in the users table in the administration

---

🛠️ **Type of Change**

- Bug fix (non-breaking change which fixes an issue)

---

🖼️ **Screenshots (if applicable)**

### DIFF
--- a/app/Http/Controllers/Admin/UserController.php
+++ b/app/Http/Controllers/Admin/UserController.php
@@ -381,7 +381,7 @@ class UserController extends Controller
      */
     public function dataTable(Request $request)
     {
-        $query = User::query()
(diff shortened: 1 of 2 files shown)


S0ly:
please add more description

1day2die:
we removed the discordID from the users table cause it wasnt given correctly from the controller to the frontend on the admin/users table. this has to be reimplemented

DominikSLK:
> we removed the discordID from the users table cause it wasnt given correctly from the controller to the frontend on the admin/users table. this has to be reimplemented

i made a change to the controller and it seems to work fine

PR #1161, opened 2025-02-09
Don't show no nodes alert while loading

DominikSLK:
💡 **Description**

Fixes no nodes alert showing while the products are loading. Fixes no nodes alert showing while the page is loading by adding d-none class

---

🛠️ **Type of Change**

- User interface (UI) improvement

---

🖼️ **Screenshots (if applicable)**

Before:
https://github.com/user-attachments/assets/c7dbbf9b-24a0-42df-85f5-9201bc8f947a

After:
https://github.com/user-attachments/assets/061357d0-b823-4a82-928a-058c313e2bc3

### DIFF
--- a/themes/default/views/servers/create.blade.php
+++ b/themes/default/views/servers/create.blade.php
@@ -152,7 +152,7 @@
                                   </template>
                                 </select>
                               </div>
-                              <div class="p-2 m-2 alert alert-danger" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0">
+                              <div class="p-2 m-2 alert alert-danger" style="display: none" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
                                 {{ __('There seem to be no nodes available for this specification. Admins have been notified. Please try again later of contact us.') }}
                               </div>
                         </div>
--- a/themes/default/views/servers/create.blade.php
+++ b/themes/default/views/servers/create.blade.php
@@ -152,8 +152,8 @@
                                   </template>
                                 </select>
                               </div>
-                              <template x-if="true">
-                                <div class="p-2 m-2 alert alert-danger" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
+                              <template x-if="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
+                                <div class="p-2 m-2 alert alert-danger">
                                   {{ __('There seem to be no nodes available for this specification. Admins have been notified. Please try again later of contact us.') }}
                                 </div>
                               </template>
--- a/themes/default/views/servers/create.blade.php
+++ b/themes/default/views/servers/create.blade.php
@@ -152,7 +152,7 @@
                                   </template>
                                 </select>
                               </div>
-                              <div class="p-2 m-2 alert alert-danger" style="display: none" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
+                              <div class="p-2 m-2 alert alert-danger d-none" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
                                 {{ __('There seem to be no nodes available for this specification. Admins have been notified. Please try again later of contact us.') }}
                               </div>
                         </div>
--- a/themes/default/views/servers/create.blade.php
+++ b/themes/default/views/servers/create.blade.php
@@ -152,9 +152,11 @@
                                   </template>
                                 </select>
                               </div>
-                              <div class="p-2 m-2 alert alert-danger d-none" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
-                                {{ __('There seem to be no nodes available for this specification. Admins have been notified. Please try again later of contact us.') }}
-                              </div>
+                              <template x-if="true">
+                                <div class="p-2 m-2 alert alert-danger" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
+                                  {{ __('There seem to be no nodes available for this specification. Admins have been notified. Please try again later of contact us.') }}
+                                </div>
+                              </template>
                         </div>
                     </div>
                 </div>


MrWeez (changes requested):
Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.

DominikSLK:
> Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.

didn't know about that

MrWeez:
> > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> 
> didn't know about that

No problem, we can't know everything.

Have you checked the changes after the last commit, is everything working?

DominikSLK:
> > > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> > 
> > 
> > didn't know about that
> 
> No problem, we can't know everything.
> 
> Have you checked the changes after the last commit, is everything working?

yes, everything is working

MrWeez:
> > > > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> > > 
> > > 
> > > didn't know about that
> > 
> > No problem, we can't know everything.
> > 
> > Have you checked the changes after the last commit, is everything working?
> 
> yes, everything is working

To clarify, is the notification still there if the free node not found?

DominikSLK:
> > > > > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> > > > 
> > > > 
> > > > didn't know about that
> > > 
> > > 
> > > No problem, we can't know everything.
> > > Have you checked the changes after the last commit, is everything working?
> > 
> > 
> > yes, everything is working
> 
> To clarify, is the notification still there if the free node not found?

forgot about that, it actually is not with the d-none class, i am not sure what to do now

MrWeez:
> > > > > > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> > > > > 
> > > > > 
> > > > > didn't know about that
> > > > 
> > > > 
> > > > No problem, we can't know everything.
> > > > Have you checked the changes after the last commit, is everything working?
> > > 
> > > 
> > > yes, everything is working
> > 
> > To clarify, is the notification still there if the free node not found?
> 
> forgot about that, it actually is not with the d-none class, i am not sure what to do now

Try adding x-cloak instead of .d-none, this should hide the element before alpinejs initialization

https://alpinejs.dev/directives/cloak

Ferks-FK:
> > > > > > > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> > > > > > 
> > > > > > 
> > > > > > didn't know about that
> > > > > 
> > > > > 
> > > > > No problem, we can't know everything.
> > > > > Have you checked the changes after the last commit, is everything working?
> > > > 
> > > > 
> > > > yes, everything is working
> > > 
> > > 
> > > To clarify, is the notification still there if the free node not found?
> > 
> > 
> > forgot about that, it actually is not with the d-none class, i am not sure what to do now
> 
> Try adding x-cloak instead of .d-none, this should hide the element before alpinejs initialization
> 
> https://alpinejs.dev/directives/cloak

Correct, but x-cloak is not defined anywhere.
It can do this in the layout of the two default themes.

DominikSLK:
> > > > > > > > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> > > > > > > 
> > > > > > > 
> > > > > > > didn't know about that
> > > > > > 
> > > > > > 
> > > > > > No problem, we can't know everything.
> > > > > > Have you checked the changes after the last commit, is everything working?
> > > > > 
> > > > > 
> > > > > yes, everything is working
> > > > 
> > > > 
> > > > To clarify, is the notification still there if the free node not found?
> > > 
> > > 
> > > forgot about that, it actually is not with the d-none class, i am not sure what to do now
> > 
> > 
> > Try adding x-cloak instead of .d-none, this should hide the element before alpinejs initialization
> > https://alpinejs.dev/directives/cloak
> 
> Correct, but x-cloak is not defined anywhere. It can do this in the layout of the two default themes.

would not be better to just use <template x-if="true"> that does not require the style?

Ferks-FK:
> > > > > > > > > Why use inline css if bootstrap provides the same thing through classes? As far as I know, .d-none works exactly the same way.
> > > > > > > > 
> > > > > > > > 
> > > > > > > > didn't know about that
> > > > > > > 
> > > > > > > 
> > > > > > > No problem, we can't know everything.
> > > > > > > Have you checked the changes after the last commit, is everything working?
> > > > > > 
> > > > > > 
> > > > > > yes, everything is working
> > > > > 
> > > > > 
> > > > > To clarify, is the notification still there if the free node not found?
> > > > 
> > > > 
> > > > forgot about that, it actually is not with the d-none class, i am not sure what to do now
> > > 
> > > 
> > > Try adding x-cloak instead of .d-none, this should hide the element before alpinejs initialization
> > > https://alpinejs.dev/directives/cloak
> > 
> > 
> > Correct, but x-cloak is not defined anywhere. It can do this in the layout of the two default themes.
> 
> would not be better to just use `<template x-if="true">` that does not require the style?

Try and see

themes/default/views/servers/create.blade.php
```diff
@@ -152,9 +152,11 @@ class="fas fa-info-circle"></i>
                                   </template>
                                 </select>
                               </div>
-                              <div class="p-2 m-2 alert alert-danger d-none" x-show="selectedProduct != null && selectedProduct != '' && locations.length == 0 && !loading">
-                                {{ __('There seem to be no nodes available for this specification. Admins have been notified. Please try again later of contact us.') }}
-                              </div>
+                              <template x-if="true">
```
Ferks-FK, line 155: You can't just pass `true` for the condition.
You must move everything that was in `x-show` into `x-if`.

Ferks-FK (changes requested):

themes/default/views/servers/create.blade.php
DominikSLK: it does work like that, but yeah it will look better

1day2die:
LGTM? @Ferks-FK

Ferks-FK:
> LGTM? @Ferks-FK

Yep, sounds good.

PR #1184, opened 2025-02-24
v1.0.7

1day2die:
´


app/Models/User.php
AGuyNamedJens, line 1: Is there a reason for this to still exist if it's pretty much discontinued and removed everywhere else?

AGuyNamedJens (commented):
Looks good to me

app/Models/User.php
1day2die: it will be removed in an later update

PR #1188, opened 2025-02-26
[WIP] fix: floating point error, switching to int values instead of float ones

Mr666dd:
Good day! I started changing the system: I changed a couple of migrations, several controllers, and 2 UI pages; I had a permission issue, so it says I changed everything except storage, but that's not true.


S0ly:
warning your dev branch seems not up to date

PR #1196, opened 2025-02-28
fix: floating point error

Mr666dd:
I started working on changing the system: credits in users is now an integer; I changed Products, and the entire panel is actively changing. For all questions, suggestions and changes, I am open to your suggestions. Thanks!

### DIFF
--- a/app/Http/Controllers/Admin/ProductController.php
+++ b/app/Http/Controllers/Admin/ProductController.php
@@ -174,7 +174,7 @@ class ProductController extends Controller
 
         $disabled = ! is_null($request->input('disabled'));
         $oomkiller = ! is_null($request->input('oom_killer'));
-        $price = intval($request->input('price'))*100;
+        $price = intval($request->input('price')) * 100;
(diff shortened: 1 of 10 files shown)


Mr666dd:
Can someone leave a review of the current code?

S0ly:
yes he seems to do, you two are doing the same error in the database migration
you need to understand where the error is coming from before trying to fix it

Mr666dd:
> > Может кто-нибудь оставить отзыв о текущем коде?
> 
> просто ради моего любопытства, вы использовали искусственный интеллект?

Such an important system, payments, does not allow the use of artificial intelligence. Who knows what he might do; I only have the handwritten code and only with my own hands

Mr666dd:
> yes he seems to do, you two are doing the same error in the database migration you need to understand where the error is coming from before trying to fix it

I don't understand what??
Dare I suggest that you first need to change the value, and then change the type in the table? Am I right?

1day2die:
fixed by ferks

PR #1209, opened 2025-03-28
fix: servers not deleting on creation failure

DominikSLK:
💡 **Description**

Fixes servers not being deleted when the creation fails on pterodactyl by removing exception throw in `PterodactylClient` since the check for response is in `ServerController`.

---

🛠️ **Type of Change**

- Bug fix (non-breaking change which fixes an issue)


S0ly:
legit, @1day2die can we merge ?

Issue #1210, opened 2025-03-29
Server limits were increased with each purchase, ignoring the limits


Ferks-FK:
### What is your request about?

title.

### Branch

main

### CtrlPanel Logs

```Shell

```

Issue #1095, opened 2025-01-05
Floating Point Errors


S0ly:
### What is your request about?

currently the panel is using floats for lots of things, this could become a very big problems
most of the banking system avoid to use floats

a refactor should be aiming to fix this issue

ethier by reworking everything (best but slow)
- Database - Store as integers (cents)
- Application logic - All calculations must use integers
- UI - Display as floats but handle all input as integers

or by doing a quick fix (quick but bad)
- bcmath on all application logic

### Branch

main

PR #1211, opened 2025-03-29
fix: floating point error

Ferks-FK:
This PR fixes the problem reported here: #1095
It also already has the fix for: #1210.

PR implements a `1000` division and multiplication factor to support currencies with 3 decimal places.

Edit:
In the commit https://github.com/Ctrlpanel-gg/panel/pull/1211/commits/7d2c1020ad643c4b6ea271328659100d61936229 the way `minimum_credits` works has been changed.
Now the old `-1` that was used to use the configuration value, becomes `null`, to avoid users putting `-1, -10, -100` thinking it would work.

Todo:
- [x] Test the gateways (Mercado Pago, Mollie, Paypal, Stripe).
- [x] Test the partner discounts.


MrWeez (commented):
Why do you divide and multiply by 1000? If we store prices in cents, then we need to use 100 or am I wrong?

MrWeez:
> Why do you divide and multiply by 1000? If we store prices in cents, then we need to use 100 or am I wrong?

Oh, my bad, I didn't know about the existence of currencies with three decimal.

1day2die:
code LGTM, will spin up a test env

S0ly (approved):
seems good to me

S0ly:
This PR seems to include breaking changes for Payments gateway 

we may want to release as 1.1.0 so show that

PR #1230, opened 2025-05-27
1.1.0. Dev

1day2die:
[Fixes]
Fixed Invoices showing the wrong price
Fixed php artisan cp:toggle:recaptcha to enable/disable recaptcha via console
Possibly fixed negative credits bug
Bug with endless-loading during server creation caused by old roles being used
Removed GLOB_BRACE in config/app.php
When cloning a product the billing period will be the one of the original 

[Added]
New discord role for Users with active Servers
custom additional notes on Invoices (praktisch für Deutsche Kleinunternehmen)
Added reason-input when admin suspends a server by hand - shown in activity logs

[Others]
Replaces all calls of env() with the corresponding config() counterpart

### DIFF
--- a/app/Http/Controllers/ProductController.php
+++ b/app/Http/Controllers/ProductController.php
@@ -18,6 +18,7 @@ use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Notification;
 use Illuminate\Support\Facades\RateLimiter;
+use App\Models\Role;
 
 class ProductController extends Controller
 {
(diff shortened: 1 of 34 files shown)


S0ly:
It should be 1.1.0 it contain breaking changes for all payment gateway

1day2die:
done

app/Http/Controllers/ProductController.php
```diff
@@ -103,10 +104,10 @@ public function getLocationsBasedOnEgg(Request $request, Egg $egg)
             // Rate limit the node full notification to 1 attempt per 30 minutes
             RateLimiter::attempt(
                 key: 'nodes-full-warning',
-                maxAttempts: 1,
+                maxAttempts: 2,
                 callback: function() {
                     // get admin role and check users
-                    $users = User::query()->where('role', '=', '1')->get();
+                    $users = User::role(1)->get();
```
MrWeez, line 110: Maybe you should link this to the "view errors" permission and not to the role? I think this will be a more correct implementation, besides, we already have such permission

1day2die: oooh thats what you meant. thats a very good idea!

MrWeez: > oooh thats what you meant. thats a very good idea!

This is what Soly has been trying to say for a long time :<

PR #1252, opened 2025-07-04
Fix create user command

Ferks-FK:
Fixes user creation via the artisan command, and adds email verification and pterodactyl ID.


Ferks-FK:
I made a mistake when checking out, another PR will be opened.

PR #1287, opened 2025-11-12
Fix Server Create // Update Composer Packages // 

1day2die:


simbabimba-dev:
@1day2die I suggest the creation of the server be first in order and then when its created on pterodactyl, create a record on ctrlpanel db

PR #1291, opened 2026-01-10
Change product change list from "created at" to price

KroZen-Dev:
This pull request makes a targeted change to the `getUpgradeOptions` function in `ServerController.php` to improve the way upgrade options are sorted. Instead of ordering products by their creation date, the products are now ordered by price in ascending order, which should provide users with a more logical and user-friendly upgrade path.

- Changed product sorting in `getUpgradeOptions` to order by `price` (ascending) instead of `created_at` for more relevant upgrade option display.


MrWeez:
Don't commit to main!!! Only dev branch

PR #1300, opened 2026-01-26
fix: check permission admin icons

supervaizir:


MrWeez:
We don't accept PRs to main branch.

app/Http/Controllers/Admin/SettingsController.php
```diff
@@ -26,6 +26,7 @@ class SettingsController extends Controller
      */
     public function index()
     {
+        $this->checkAnyPermission([self::ICON_PERMISSION]);
```
MrWeez, line 29: Why `ICON_PERMISSION` and not `SETTINGS_PERMISSIONS` as it should be? Either you haven't fully learned our codebase, or it's an inappropriate suggestion from an AI that doesn't know the structure of the project.
> *Small note: We are not against using AI for coding, but it should be used as an assistant, under strict human control.*

Reference: https://github.com/Ctrlpanel-gg/panel/blob/development/app/Constants/PermissionGroups.php#L32-L65

MrWeez (changes requested):

PR #1275, opened 2025-09-25
🔧 Multiple Bug Fixes & Feature Enhancements

simbabimba-dev:
## Description

This pull request addresses several critical bugs and implements requested features to improve CtrlPanel's stability, security, and usability. The changes include checkbox persistence fixes, unlimited resource configuration, enhanced coupon logic, error handling improvements, and credit system refinements for the referrals. I have made this PR as a single one because the changes are relatively small and interconnected, and it would be inefficient to create separate PRs for each issue and made it so that you can cherry-pick the commits you want if you don't want all of them.

---

## 🔗 Linked Issues

-   Fixes [#1272 - OOM Killer checkbox not saving state](https://github.com/Ctrlpanel-gg/panel/issues/1272)
-   Fixes [#1241 - Allow 0 on products for unlimited specifications](https://github.com/Ctrlpanel-gg/panel/issues/1241)
-   Fixes [#1240 - Allow unlimited coupon uses](https://github.com/Ctrlpanel-gg/panel/issues/1240)
-   Fixes [#1255 - Error 500 on overview page if /.git/HEAD missing](https://github.com/Ctrlpanel-gg/panel/issues/1255)
-   Fixes a small part of [[#1248 - Problems related to displaying/formatting incorrectly formatted coins](https://github.com/Ctrlpanel-gg/panel/issues/1248)] (related to the initial credits given to the 1st admin user when the panel is installed)

---

## 📋 Summary of Changes

| Component             | Description                                                                                     | Issue                                                      | Commit                                                                                                                                                                                                                    | Status |
| --------------------- | ----------------------------------------------------------------------------------------------- | ---------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------ |
| **User Settings**     | Fix values from the user settings, currencycast cache issue, logic moved to settings controller | Self-discovered bugs                                       | [`24538f98`](https://github.com/LakshmiBhaskarPVL/cpgg/commit/24538f98), [`c1eb6858`](https://github.com/LakshmiBhaskarPVL/cpgg/commit/c1eb6858), [`06910690`](https://github.com/LakshmiBhaskarPVL/cpgg/commit/06910690) | ✅     |
[...]

PR #1306, opened 2026-02-02
Prevent negative credit balance

### LINKED ISSUE #1302: Change how "Minimum Credits" works for products

It was a good option, with monthly billing, based on hourly charges. But since 1.0, when we have dynamic billing, so this function can create a problem of negative balance.

### Why?

User has 500 credits
Product costs 1000 credits, and has monthly billing
Product "Minimum Credits" is set to 500

If user creates server, system will take 1000 credits from him, to pay for a full month.

Now user has -500 credits

### Branch

main

### Solution idea

We should either remove this feature, or do validation so that the "Minimum Credits" cannot be less than the price of the product.

Also, we should remove "Min Credits To Make Server" from the settings, as this feature no longer makes sense.

### CtrlPanel Logs

```Shell

```

### Additional Info

From discord development discussion ([original message](https://discord.com/channels/787829714483019826/851144461361872957/1467937497232314533)):
1. Remove global min credits (From settings, controllers and database)
2. Create validation, to prevent per products Min credits to be less than the product price (frontend and backend)
3. Carefully migrate existing products Min credits
    - If the current value was default -> set it to product price
    - If the current value is less than the product price -> set it to the product price
    - If the current value is greater or equals to the product price -> leave as it is

KroZen-Dev:
This pull request refactors the handling of minimum credits required to create a server, simplifying the logic to always use the product's price as the minimum. It removes the global `min_credits_to_make_server` user setting and updates validation, UI, and translations accordingly.

**Product minimum credits and validation changes:**
- The `minimum_credits` field for products now must be greater than or equal to the product's `price`, enforced in both the `store` and `update` methods in `ProductController` (`app/Http/Controllers/Admin/ProductController.php`). [[1]](diffhunk://#diff-18b18dc227920b7cef275dbcf6e7a66af2b92dc41611e6f6622df3437323bd04L88-R88) [[2]](diffhunk://#diff-18b18dc227920b7cef275dbcf6e7a66af2b92dc41611e6f6622df3437323bd04L167-R167)
- The server creation requirement now checks only if the user's credits are at least the product's price, removing fallback to a global minimum credits value (`app/Http/Controllers/ServerController.php`).

**Removal of global minimum credits setting:**
- The `min_credits_to_make_server` property and all related validation, casting, and input data are removed from `UserSettings` (`app/Settings/UserSettings.php`). [[1]](diffhunk://#diff-8343900f0e487abc3e94fe3cc8918777e1a0f1f5c38cfecb21d3d1cafcdd1d75L18) [[2]](diffhunk://#diff-8343900f0e487abc3e94fe3cc8918777e1a0f1f5c38cfecb21d3d1cafcdd1d75L39) [[3]](diffhunk://#diff-8343900f0e487abc3e94fe3cc8918777e1a0f1f5c38cfecb21d3d1cafcdd1d75L56) [[4]](diffhunk://#diff-8343900f0e487abc3e94fe3cc8918777e1a0f1f5c38cfecb21d3d1cafcdd1d75L108-L113)
- The variable `min_credits_to_make_server` is removed from the server creation view context (`ServerController.php`).

**UI and messaging updates:**
- The server creation form and product forms now display "Price" instead of "Minimum Credits" and update tooltips to clarify that the minimum cannot be less than the product price (`themes/default/views/servers/create.blade.php`, `themes/default/views/admin/products/create.blade.php`, `themes/default/views/admin/products/edit.blade.php`). [[1]](diffhunk://#diff-4cdbff7ebff596a2d8336e1361d7df5be7637614f7ab5f72cb835208ee21a4e1L277-R279) [[2]](diffhunk://#diff-8fcfbc68ad119d9995b7b36418466cb62934552fd1e9635764f3311ad3102382L170-R170) [[3]](diffhunk://#diff-5a79d48bbb36bbd829330e2824675242ed3323f4909db28ab3c9218321380ae3L180-R180)
- Button enable/disable logic and buy-more-credits prompts now depend solely on the product's price, not both price and minimum credits (`themes/default/views/servers/create.blade.php`).

🔗 Linked Issues:

Fixes #1302


app/Http/Controllers/Admin/ProductController.php
```diff
@@ -86,8 +86,8 @@ public function store(Request $request)
             'cpu' => 'required|numeric|max:1000000|min:0',
             'swap' => ['required', $this->getSwapValidator()],
             'description' => 'required|string|max:191',
-            'disk' => 'required|numeric|max:1000000|min:0',
-            'minimum_credits' => 'nullable|numeric|max:1000000',
+            'disk' => 'required|numeric|max:1000000|min:5',
```
MrWeez: Why?

app/Settings/UserSettings.php
```diff
@@ -30,15 +29,14 @@ public static function group(): string
...
-    // public static function casts(): array
-    // {
-    //     return [
-    //         'credits_reward_after_verify_discord' => CurrencyCast::class,
-    //         'credits_reward_after_verify_email' => CurrencyCast::class,
-    //         'initial_credits' => CurrencyCast::class,
-    //         'min_credits_to_make_server' => CurrencyCast::class,
-    //     ];
-    // }
+    public static function casts(): array
```
MrWeez: I don't think this is related to PR. Also, why?

app/Http/Controllers/ServerController.php
```diff
@@ -187,9 +186,7 @@ private function validateProductRequirements(Product $product, Request $request)
             return __('You can not create any more Servers with this product!');
         }
 
-        $minCredits = $product->minimum_credits ?: $this->userSettings->min_credits_to_make_server;
-
-        if ($user->credits < $minCredits) {
```
MrWeez, line 192: We still want to have minimum credits option, so you don't need to remove that part

app/Http/Controllers/ServerController.php
```diff
@@ -187,9 +186,7 @@ private function validateProductRequirements(Product $product, Request $request)
...
+        if ($user->credits < $product->price) {
```
MrWeez:
```suggestion
        $minCredits = $product->minimum_credits;

        if ($user->credits < $minCredits) {
```

app/Settings/UserSettings.php
```diff
@@ -30,15 +29,14 @@ public static function group(): string
...
+    {
+        return [
+            'credits_reward_after_verify_discord' => CurrencyCast::class,
+            'credits_reward_after_verify_email' => CurrencyCast::class,
+            'initial_credits' => CurrencyCast::class,
+        ];
+    }
```
MrWeez, line 39: Why are you overwriting changes made by #1275 ([see](https://github.com/Ctrlpanel-gg/panel/pull/1275/changes#diff-8343900f0e487abc3e94fe3cc8918777e1a0f1f5c38cfecb21d3d1cafcdd1d75R33-R41))

themes/default/views/admin/products/create.blade.php
```diff
@@ -164,22 +164,22 @@ class="form-control @error('price') is-invalid @enderror"
...
+												value="{{ old('minimum_credits', (isset($product) && $product->minimum_credits) ? Currency::formatForForm($product->minimum_credits) : null) }}"
+												id="minimum_credits" name="minimum_credits" type="number"
+												class="form-control @error('minimum_credits') is-invalid @enderror">
+											@error('minimum_credits')
+												<div class="invalid-feedback">
+													{{ $message }}
+												</div>
+											@enderror
+										</div>
+									</div>
```
MrWeez, line 182: Please configure your editor to use four spaces instead of tabs. It would help keep the formatting consistent.

themes/default/views/servers/create.blade.php
```diff
@@ -271,9 +271,9 @@ class="fas fa-info-circle"></i>
                                                 </li>
                                                 <li class="d-flex justify-content-between">
                                                     <span class="d-inline-block"><i class="fa fa-coins"></i>
-                                                        {{ __('Minimum') }} {{ $credits_display_name }}</span>
+                                                        {{ __('Price') }} {{ $credits_display_name }}</span>
```
MrWeez, line 274: We want to keep the "Minimum credits", not remove. Just change it's behavour

themes/default/views/servers/create.blade.php
```diff
@@ -293,28 +293,31 @@ class="fas fa-info-circle"></i>
...
+                                                submitClicked ? 'disabled' : ''"
+                                            class="mt-2 btn btn-primary btn-block" @click="setProduct(product.id);"
+                                            x-text="product.doesNotFit == true
+                                                ? '{{ __('Server cant fit on this Location') }}'
+                                                : (product.servers_count >= product.serverlimit && product.serverlimit != 0
+                                                    ? '{{ __('Max. Servers with configuration reached') }}'
+                                                    : (product.price > user.credits
+                                                        ? '{{ __('Not enough') }} {{ $credits_display_name }}!'
+                                                        : '{{ __('Create server') }}'))">
+                                        </button>
```
MrWeez, line 312: Same as above. Dont need to remove it completely, just backend check so it can't less than product price itself

themes/default/views/servers/create.blade.php
```diff
@@ -293,28 +293,31 @@ class="fas fa-info-circle"></i>
...
-                                                    </button>
-                                                </a>
-                                            </template>
+                                        <template x-if="product.price > user.credits">
+                                            <a href="{{ route('store.index') }}">
+                                                <button type="button" class="mt-2 btn btn-warning btn-block">
+                                                    {{ __('Buy more') }} {{ $credits_display_name }}
+                                                </button>
+                                            </a>
+                                        </template>
```
MrWeez, line 320: Same as above.

MrWeez (changes requested):
1. I see that you removed Default min credits from settings, but you haven't removed it from the database.
2. We need to migrate existing products. And BIG note: only products, that have "Min credits" less than the product price, and don't change anything if the Min credits is greater and equals
3. Overwriting changes made by another PR (See review comments)
4. You're partially removed Minimum credits. Why? We only need to remove default value, and change per product, so it can't be less than the product price
5. Since I'm not the backend dev, I can't tell exactly, but I think there is more places that need to be changed.

MrWeez:
If you need more context, see [this](https://discord.com/channels/787829714483019826/851144461361872957/1467937497232314533) message in discord development discussion channel

app/Http/Controllers/ServerController.php
KroZen-Dev: I'm a bit confused. Are we ignoring the issue description now?

You explicitly wrote: "Remove global min credits (From settings, controllers and database)" and "Remove 'Min Credits To Make Server' from the settings, as this feature no longer makes sense."

By asking me to keep these lines, you are:
Contradicting your own "Solution idea"
Keeping the "problem of negative balance" (the 500 vs 1000 credits example) that you specifically asked to fix
Forcing the code to use a global setting that, according to the task, should no longer exist in the database

MrWeez: > I'm a bit confused. Are we ignoring the issue description now?
> 
> You explicitly wrote: "Remove global min credits (From settings, controllers and database)" and "Remove 'Min Credits To Make Server' from the settings, as this feature no longer makes sense."
> 
> By asking me to keep these lines, you are: Contradicting your own "Solution idea" Keeping the "problem of negative balance" (the 500 vs 1000 credits example) that you specifically asked to fix Forcing the code to use a global setting that, according to the task, should no longer exist in the database

> "Remove global min credits (From settings, controllers and database)"

Global value is one that is not specific to a product, which is the default value. Global = default

> "Remove 'Min Credits To Make Server' from the settings, as this feature no longer makes sense."

This is how default min credits called in the settings

I'm sorry if I created confusion.

app/Http/Controllers/Admin/ProductController.php
1day2die: 0 is actually usefull as 0 = unlimited

MrWeez: > 0 is actually usefull as 0 = unlimited

Some people want to do unlimited resources. Also, this is probably another case where Krozen make a revert of some changes that have already been made in development

MrWeez:
[Ref](https://discord.com/channels/787829714483019826/867691623479246878/1468202114332495979)

Thanks for your work anyway.

PR #1309, opened 2026-02-08
Add the option to choose a dynamic billing period.

### LINKED ISSUE #1285: Add Billing Period Selection (Hourly / Daily / Monthly) in Deployment Menu
### Feature Description

Hello CTRLPanel team 👋

Currently, the billing period (per hour, per day, per month, etc.) is configured directly inside the shop packages, which makes it rigid and visually unappealing when users deploy servers.
I’d like to suggest an improvement that allows users to choose their billing period dynamically during deployment, rather than having this hardcoded inside each package.
Please see the attached screenshots for reference.

---

**1. Current Behavior**
Billing cycle (hourly, daily, monthly) is defined within the shop package configuration.
This makes it necessary to create multiple separate packages for the same server configuration (e.g., “1024MB Hourly” and “1024MB Monthly”).
It also causes UI clutter and confusion for users when browsing available resources.

**2. Requested Feature: Select Billing Period During Deployment**
Goal: Allow the customer to select their preferred billing cycle when deploying a new server for example:

- Per Hour
- Per Day
- Per Month

**3. Flexible Package System**
Goal: Allow a single shop package to support multiple billing cycles.

Currently, admins must create one package per cycle.
Instead, each package could define all supported cycles with associated prices, for example:

Billing Type	
Hourly	
1.00 Credits	
1024MB per hour

Daily	
15.00 Credits	
1024MB per day

Monthly	
300.00 Credits	
1024MB per month

Then, when a user deploys the server, they can select their preferred billing option from that same package.
This would reduce package duplication, simplify management, and give users more flexibility.

---

**Suggested Implementation**
Add a new field in the Server Configuration menu that allows the user to choose the desired billing period before confirming deployment.

Example placement:
Extra dropdown next to the “Billing Priority” dropdown, add:

Billing Period:
[ Per Hour | Per Day | Per Month ]

Automatically update the pricing preview in real time based on the selection.

**Benefits**
✅ Cleaner and more modern deployment UI
✅ No need to duplicate packages for each billing cycle
✅ Users have more control and transparency over their costs
✅ Easier to maintain pricing in one central package configuration

**Summary**
Requested Features:
- Add billing period selector (hour/day/month) in deployment menu
- Allow multiple billing cycles to exist within one package
- Update pricing dynamically when user changes cycle
- Improve overall package management and presentation

**In short:**
Let users choose how they want to be billed (hourly, daily, monthly) when deploying and let admins define multiple billing options per package instead of creating duplicates. This would significantly improve usability and flexibility for both admins and customers.

Ferks-FK:
implements #1285.

### DIFF
--- a/app/Console/Commands/ChargeServers.php
+++ b/app/Console/Commands/ChargeServers.php
@@ -61,19 +61,8 @@ class ChargeServers extends Command
                     $product = $server->product;
                     /** @var User $user */
                     $user = $server->user;
-
-                    $billing_period = BillingPeriod::from($server->billing_period->value ?? $product->default_billing_period->value);
-
-                    // check if server is due to be charged by comparing its last_billed date with the current date and the billing period
-                    $newBillingDate = match($billing_period) {
-                        BillingPeriod::ANNUALLY => Carbon::parse($server->last_billed)->addYear(),
-                        BillingPeriod::HALF_ANNUALLY => Carbon::parse($server->last_billed)->addMonths(6),
-                        BillingPeriod::QUARTERLY => Carbon::parse($server->last_billed)->addMonths(3),
-                        BillingPeriod::MONTHLY => Carbon::parse($server->last_billed)->addMonth(),
-                        BillingPeriod::WEEKLY => Carbon::parse($server->last_billed)->addWeek(),
-                        BillingPeriod::DAILY => Carbon::parse($server->last_billed)->addDay(),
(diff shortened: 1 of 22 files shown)


KroZen-Dev:
Hi! If I reviewed the pull request correctly, there's no option to enable/disable the ability to select a custom payment period. For example, if an administrator wants to create one plan with a fixed payment period and a second plan with custom periods, I think an on/off option would be very useful

1day2die:
@Ferks-FK  status?

PR #1311, opened 2026-03-06
Add ability to revoke server cancellation

KroZen-Dev:
## Problem
Currently, once a server is canceled, the Cancel button becomes inactive and users cannot revert the cancellation. This prevents restoring a server if the cancellation was accidental.

## Solution
- Introduce a PATCH route to uncancel a server
- Add uncancel() method in ServerController to restore canceled servers
- Update the server index Blade template to show conditional Cancel (yellow) or Restore (green) buttons
- Add SweetAlert2 confirmation dialog for uncancel action
- Ensure users can restore canceled servers before the billing period ends

## Changes
- routes/web.php
- app/Http/Controllers/ServerController.php
- themes/default/views/servers/index.blade.php

## Features
- Active servers show a yellow Cancel button
- Canceled servers show a green Restore button
- Users receive proper feedback on server restoration

### DIFF
--- a/app/Http/Controllers/ServerController.php
+++ b/app/Http/Controllers/ServerController.php
@@ -388,6 +388,22 @@ class ServerController extends Controller
         }
     }
 
+    public function uncancel(Server $server): RedirectResponse
+    {
+        if ($server->user_id !== Auth::id()) {
+            return back()->with('error', __('This is not your Server!'));
+        }
+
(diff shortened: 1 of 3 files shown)


1day2die:
Thanks for this!

please add what copilot suggested. Checking for current suspension is a good idea i think

grafik

PR #1313, opened 2026-03-10
Remove global minimum‑credits setting and enforce per‑product minimums

### LINKED ISSUE #1302: Change how "Minimum Credits" works for products

simbabimba-dev:
## **Summary**
- Removes the global “min credits” setting and enforces per‑product minimums.
- Adds a migration to normalize existing product `minimum_credits` values so they follow the new rules:
  - Default (`-1`) → set to `price`
  - Any `minimum_credits` < `price` → set to `price`
  - Values ≥ `price` remain unchanged
---

## **What changed**
- **Validation:** `minimum_credits` must be >= `price` on product create/update. (ProductController.php, API requests)
- **Data migration:** `2026_02_02_175351_migrate_product_minimum_credits_values.php` normalises old values.
- **Settings cleanup:** removed stored global setting; migration 2026_02_02_182405_remove_min_credits_setting.php deletes `user.min_credits_to_make_server` and property removed from `UserSettings.php`.
- **Server logic:** creation/unsuspend routines now use product price when minimum is unset/default. (ServerController.php, UnsuspendServers.php)
- **Model/UX:** `Product` accessor and blade views display/use effective minimum; updated UI text and language entries. (Product.php, `themes/default/views/...`, en.json)
- **Misc:** validation added in API request classes; product resource returns effective minimum; transaction/locking added to listener.
- Note: This is a data migration too ; `down()` restores the backed‑up column.

---

## **Upgrade / deploy notes**
1. Backup database before deploying (migration alters existing `products` data).
2. Run migrations:  
   ```bash
   php artisan migrate
   ```
3. Test product editing, server creation, and suspension/unsuspension in staging.

---

### **Related**
- Fixes/addresses: #1302


simbabimba-dev (commented):
LGTM

PR #1316, opened 2026-03-22
feat(provisioning): robust server creation + reconciliation safeguards

simbabimba-dev:
## Problem
Server provisioning was previously handled as a single-step process with no clear state tracking or recovery path. Failures during API calls or database operations could lead to inconsistent states, such as users being charged without receiving a server or servers being created without proper tracking.

## Solution
- Introduce explicit server lifecycle states (`provisioning`, `active`, `failed`, `pending_reconciliation`)
- Implement a dedicated service to handle server creation, credit reservation, and failure recovery
- Add a reconciliation job to resolve uncertain or partial failures
- Introduce atomic credit handling to prevent race conditions and ensure safe refunds
- Improve verification of remote server state via external ID lookup

## Features
- Servers are created with a tracked provisioning lifecycle
- Users are only charged when provisioning is confirmed or safely handled
- Failed or uncertain provisioning attempts are automatically reconciled
- Remote server state is verified to prevent orphaned or duplicate servers
- Credit operations are atomic and protected against race conditions

### DIFF
--- a/app/Classes/PterodactylClient.php
+++ b/app/Classes/PterodactylClient.php
@@ -302,6 +302,22 @@ class PterodactylClient
         }
     }
 
+    /**
+     * Get a server by external_id on Pterodactyl.
+     *
+     * @param string $externalId
+     * @return \Illuminate\Http\Client\Response
+     * @throws Exception
+     */
+    public function getServerByExternalId(string $externalId)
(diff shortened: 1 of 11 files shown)


1day2die:
LGTM

PR #1319, opened 2026-03-30
Server balancer: auto-select node with free allocations (#1310)

### LINKED ISSUE #1310: Server balancer option
### Feature Description

Hi, I have a problem, I set allocation limit, and want to create a new server but on selected node no place, but other nodes have

### Solution Idea

add flag for case when other node is availible for automatic server creation on available node 

### Additional Info

_No response_

simbabimba-dev:
## Problem
Server provisioning could fail when the selected node had no free IP/port allocations even though other nodes had available allocations. The previous selection logic validated only memory and disk resources, causing provisioning errors, ambiguous validation messages, and failed deployments when allocations were exhausted on the chosen node.

## Solution
- Validate IP/port allocation availability in node selection in addition to memory and disk.
- Add `findAvailableNodeWithAllocation` to locate nodes that have free allocations and satisfy resource constraints.
- Update `ServerController` to use `findAvailableNodeWithAllocation`; `ServerCreationService` adds the helper but retains a `findAvailableNode` + `getFreeAllocationId` flow (functionally allocation-aware, but usage differs).
- Surface explicit validation errors when no node with free allocations exists to avoid silent failures.
- Note: This commit does not add an explicit "server balancer" configuration flag; the fallback to another node with allocations is unconditional within the searched location (the helper provides an integration point for future balancer behavior).

## Features
- Node selection checks allocations as well as memory/disk.
- Automatic fallback to the first available node with a free allocation within the selected location (no balancer toggle implemented here).
- Clear validation error and failure handling when allocations are exhausted.
- Reduced provisioning failures and allocation conflicts during server creation.
- Adds `findAvailableNodeWithAllocation` as an integration point for the Server Balancer issue (#1310).

### DIFF
--- a/app/Http/Controllers/ServerController.php
+++ b/app/Http/Controllers/ServerController.php
@@ -174,10 +174,10 @@ class ServerController extends Controller
     private function validateProductRequirements(Product $product, Request $request): string|bool
     {
         $location = $request->input('location');
-        $availableNode = $this->findAvailableNode($location, $product);
+        $nodeAllocation = $this->findAvailableNodeWithAllocation($location, $product);
 
-        if (!$availableNode) {
-            return __("The chosen location doesn't have the required memory or disk left to allocate this product.");
+        if (!$nodeAllocation) {
+            return __("The chosen location doesn't have the required memory, disk, or free allocations to allocate this product.");
         }
 
         $user = Auth::user();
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/ServerController.php
```diff
@@ -174,10 +174,10 @@ private function validateServerCreation(Request $request): ?RedirectResponse
     private function validateProductRequirements(Product $product, Request $request): string|bool
     {
         $location = $request->input('location');
-        $availableNode = $this->findAvailableNode($location, $product);
+        $nodeAllocation = $this->findAvailableNodeWithAllocation($location, $product);
 
-        if (!$availableNode) {
-            return __("The chosen location doesn't have the required memory or disk left to allocate this product.");
+        if (!$nodeAllocation) {
+            return __("The chosen location doesn't have the required memory, disk, or free allocations to allocate this product.");
```
MrWeez: I think it will sound better if you change that to smth like "the selected location does not have the required memory, disk, or overloaded"

MrWeez:
Except that one thing, looks great to me

simbabimba-dev:
oki, will fix that and do it rn

MrWeez (approved):

PR #1338, opened 2026-04-28
fixes for v1.2.1

simbabimba-dev:
Fixes bugs we found during public deployment of v1.2
Pretty self explanatory
issues fixed
- mariadb/mysql incompatible migrations
- search using discord id
- port consideration for pterodactyl in installer
- default role perms overwritten by seeders

### DIFF
--- a/app/Http/Controllers/Admin/UserController.php
+++ b/app/Http/Controllers/Admin/UserController.php
@@ -151,7 +151,7 @@ class UserController extends Controller
         $this->checkPermission(self::READ_PERMISSION);
 
         $users = QueryBuilder::for(User::query())
-            ->allowedFilters(['id', 'name', 'pterodactyl_id', 'email'])
+            ->allowedFilters(['id', 'name', 'pterodactyl_id', 'email','discordUser.id'])
             ->paginate(25);
 
         if ($request->query('user_id')) {
--- a/app/Http/Controllers/Admin/UserController.php
+++ b/app/Http/Controllers/Admin/UserController.php
@@ -151,8 +151,7 @@ class UserController extends Controller
         $this->checkPermission(self::READ_PERMISSION);
 
         $users = QueryBuilder::for(User::query())
-            ->with('discordUser')
-            ->allowedFilters(['id', 'name', 'pterodactyl_id', 'email', 'discordUser.id'])
+            ->allowedFilters(['id', 'name', 'pterodactyl_id', 'email'])
             ->paginate(25);
 
         if ($request->query('user_id')) {
--- a/app/Http/Controllers/Admin/UserController.php
+++ b/app/Http/Controllers/Admin/UserController.php
@@ -151,7 +151,8 @@ class UserController extends Controller
         $this->checkPermission(self::READ_PERMISSION);
 
         $users = QueryBuilder::for(User::query())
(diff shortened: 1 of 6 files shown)


simbabimba-dev:
I have read and agree to the CLA.

app/Http/Controllers/Admin/UserController.php
```diff
@@ -151,7 +151,8 @@ public function json(Request $request)
         $this->checkPermission(self::READ_PERMISSION);
 
         $users = QueryBuilder::for(User::query())
-            ->allowedFilters(['id', 'name', 'pterodactyl_id', 'email'])
+            ->with('discordUser')
+            ->allowedFilters(['id', 'name', 'pterodactyl_id', 'email', 'discordUser.id'])
```
MrWeez: You don't need to update admin UserController for API

app/Http/Controllers/Api/UserController.php
```diff
@@ -56,6 +56,7 @@ public function __construct(PterodactylSettings $ptero_settings, ReferralSetting
     public function index(Request $request)
     {
         $users = QueryBuilder::for(User::class)
+            ->with('discordUser')
```
MrWeez: Do not include discordUser, otherwise it will be returned always, even without ?include=discordUser

app/Http/Controllers/Api/UserController.php
```diff
@@ -77,6 +78,7 @@ public function index(Request $request)
     public function show(Request $request, int $userId)
     {
         $user = QueryBuilder::for(User::class)
+            ->with('discordUser')
```
MrWeez: Same as line 59. Do not include discordUser, otherwise it will be returned always, even without ?include=discordUser

MrWeez (changes requested):

app/Http/Controllers/Admin/UserController.php
simbabimba-dev: wouldnt that prevent admins from searching for a user using a discord id?
in the panel?

MrWeez: > wouldnt that prevent admins from searching for a user using a discord id? in the panel?

Fair, but in this case you don't need `with`. Because filter automatically does this

simbabimba-dev: forgot about that, srry

MrWeez:
So, you decided to remove ability to search on admin ui using discord id?

simbabimba-dev:
sorry, i thought you were referring to the entire change. 
will fix it rn
image

MrWeez:
> sorry, i thought you were referring to the entire change. will fix it rn image

This was for the initial change, I didn't understand that you want to add a search by id, including for the admin ui, and not just the api

MrWeez (approved):
LGTM

MrWeez:
@simbabimba-dev is this PR ready to be merged? Just one this, as we discussed, you don't need discord id filter for admin user list, since it uses datatable discord_id filed

simbabimba-dev:
> @simbabimba-dev is this PR ready to be merged? Just one this, as we discussed, you don't need discord id filter for admin user list, since it uses datatable discord_id filed

PR is ready for merge

MrWeez (approved):

PR #1355, opened 2026-06-16
Add possibility to cancel server cancellation

KroZen-Dev:
This pull request improves the server uncancel (restore) workflow by introducing proper JSON response handling for API/AJAX requests and updating the UI logic across themes to fully support asynchronous operations and error states.

**Backend: Improved Response Handling & Security**

* The `uncancel` method in `ServerController` now checks for JSON expectations and returns a `204 No Content` response on success, or a `500 JSON` error on failure.
* For standard web requests, the catch block error message was updated to a secure, translatable string, removing direct exception message exposure to the end-user.

**Frontend: Enhanced Uncancel Workflow & Theme Consistency**

* Implemented the `handleServerUncancel` JavaScript function sending proper `Accept: application/json` and `X-Requested-With` headers to trigger the new backend JSON logic.
* Fixed a critical JS bug in `handleServerCancel` where the `response` variable was undefined in the callback.
* Added a SweetAlert confirmation/handling flow for the uncancel action in the default theme list (`index.blade.php`), ensuring seamless page reloads on success and explicit error alerts on failure.
* Fixed a `500 RouteNotFoundException` on the server settings page in the Phoenix theme by ensuring the template aligns with the defined routing structure.

### DIFF
--- a/app/Http/Controllers/ServerController.php
+++ b/app/Http/Controllers/ServerController.php
@@ -370,6 +370,33 @@ class ServerController extends Controller
         }
     }
 
+    public function uncancel(Server $server): \Illuminate\Http\Response|RedirectResponse
+    {
+        if ($server->user_id !== Auth::id()) {
+            return back()->with('error', __('This is not your Server!'));
+        }
+
+        try {
+            $server->update(['canceled' => null]);
+
+            if (request()->expectsJson()) {
+                return response()->noContent();
+            }
+
+            return redirect()->route('servers.index')
+                ->with('success', __('Server cancellation has been revoked'));
+        } catch (Exception $e) {
+            report($e);
+
+            if (request()->expectsJson()) {
+                return response()->json(['error' => __('Server cancellation revoke failed')], 500);
+            }
+
+            return redirect()->route('servers.index')
+                ->with('error', __('Server cancellation revoke failed'));
+        }
+    }
+
     public function show(Server $server): \Illuminate\View\View
     {
         if ($server->user_id !== Auth::id()) {
--- a/app/Http/Controllers/ServerController.php
+++ b/app/Http/Controllers/ServerController.php
@@ -21,9 +21,10 @@ use App\Enums\BillingPriority;
 use App\Settings\GeneralSettings;
 use Exception;
 use Illuminate\Database\Eloquent\Builder;
-use Illuminate\Http\Client\Response;
+use Illuminate\Http\Client\Response as ClientResponse;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
+use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Log;
@@ -354,7 +355,7 @@ class ServerController extends Controller
         Cache::forget('user_credits_left:' . $server->user_id);
     }
 
-    public function cancel(Server $server): \Illuminate\Http\Response|RedirectResponse
+    public function cancel(Server $server): Response|RedirectResponse
     {
         if ($server->user_id !== Auth::id()) {
             return back()->with('error', __('This is not your Server!'));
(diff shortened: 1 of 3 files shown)


KroZen-Dev:
I have read and agree to the CLA.

routes/web.php
```diff
@@ -65,6 +65,7 @@
     Route::get('notifications/readAll', [NotificationController::class, 'readAll'])->name('notifications.readAll');
     Route::resource('notifications', NotificationController::class);
     Route::patch('/servers/cancel/{server}', [ServerController::class, 'cancel'])->name('servers.cancel');
+    Route::patch('/servers/{server}/uncancel', [App\Http\Controllers\ServerController::class, 'uncancel'])->name('servers.uncancel');
```
MrWeez: 1. Why are you not following existing route naming? We have `/servers/cancel/:id`, so why are you using `/servers/:id/uncancel`. But at the same time I can agree that `/servers/:id/:action` format is better from hierarchy point, so if you update old route to use same format it will be ok.
2. You're using full namespace while we already have ServerController class imported on L27

themes/default/views/servers/index.blade.php
MrWeez, line 1: Why are you removed all comments? They can be useful. Leaving comments in the code is a good practice so that feature devs can easily understand code/function meaning without reading it

MrWeez (changes requested):

themes/default/views/servers/index.blade.php
KroZen-Dev: It's very rare for contributors to leave comments in code. You do not write such remarks to them, so, please do not write to me too

MrWeez: I'm not asking you to leave comments, I'm asking you not to delete existing ones without reason

routes/web.php
KroZen-Dev: Here's exactly the same style: /servers/{server}/billing_priority

themes/default/views/servers/index.blade.php
KroZen-Dev: Please tell me, is this comment in the right place? ("delete server" in the logic of the cancel button)

KroZen-Dev: As for the full namespace of the controller, I agree, I will correct for a short class

MrWeez: > Please tell me, is this comment in the right place? ("delete server" in the logic of the cancel button)
> 
> 

I agree that the comment is incorrect, and was probably just copied from the delete method and forgotten to be renamed. But is that a reason to delete it? Why not just fix it if you notice a typo instead of deleting it?

routes/web.php
MrWeez: > Here's exactly the same style: /servers/{server}/billing_priority
> 
> 

I'm not arguing that you chose a more correct format, but we're talking about uncancel route that does opposite of cancel so logically new route should follow its naming pattern OR old one should be adjusted to have same naming pattern as new and other routes.

themes/default/views/servers/index.blade.php
KroZen-Dev: You mentioned not deleting comments without a reason. The reason here is that the comment was completely wrong and misleading (stating "Delete" instead of "Cancel")

​The code fetch(route('servers.cancel', ...)) is already 100% self-explanatory and doesn't need a comment to explain what it does. Keeping obvious or broken comments just litters the codebase

app/Http/Controllers/ServerController.php
```diff
@@ -354,19 +354,57 @@ private function handleServerDeletion(Server $server): void
         Cache::forget('user_credits_left:' . $server->user_id);
     }
 
-    public function cancel(Server $server): RedirectResponse
+    public function cancel(Server $server): \Illuminate\Http\Response|RedirectResponse
```
MrWeez: Inline FQCN instead of a proper import

MrWeez (changes requested):
1. No check of current server state before uncancel. If the server isn't currently canceled (duplicate click / race between two tabs), `uncancel()` just silently sets `canceled = null` again. Not harmful, but also isn't good
2. Naming `uncancel` also stands out. It's a negation of a verb rather than an action verb itself, so it worth changing the method/route/name to something like `resume`. `servers.resume`, `ServerController::resume()` reads and sound way better

app/Http/Controllers/ServerController.php
```diff
@@ -21,9 +21,10 @@
 use App\Settings\GeneralSettings;
 use Exception;
 use Illuminate\Database\Eloquent\Builder;
-use Illuminate\Http\Client\Response;
+use Illuminate\Http\Client\Response as ClientResponse;
```
MrWeez: This import isn't used anywhere in this file, so could be removed completely instead of importing it under different name

MrWeez (commented):
Except this one small thing everything is ok.

Special thanks for changing `servers.cancel` route, which was the only one that did not comply with RESTful standards in servers related part.

PR #1356, opened 2026-07-02
Fix server limit for product upgrades/downgrades

KroZen-Dev:
## Description

Fixed server limit feature for product upgrades/downgrades. When a product has a serverlimit set (non-zero), users was able upgrade/downgrade to the product that already reached its limit for user.

### Changes:

* Added server limit validation when listing available upgrade products
* Added server limit validation when upgrading a server to a different product
* Added server limit validation in the `validateUpgrade` method

## Type of Change

* [x] Bug fix

## Testing

* [x] Tested creating a server when product limit is reached (should show error)
* [x] Tested upgrading to a product that has reached its limit (should show error)
* [x] Tested that products with `serverlimit = 0` (unlimited) work as before
* [x] Tested that upgrading to the same product doesn't trigger limit check
* [x] Verified upgrade product list correctly marks limited products as "doesNotFit"

## AI Assistance

* [ ] I used AI tools to assist with this contribution

## Checklist

* [x] My PR targets the development branch
* [x] Commit messages follow Conventional Commits
* [x] I have reviewed my own code
* [x] I have tested all affected functionality
* [x] No new warnings or errors introduced

### DIFF
--- a/app/Http/Controllers/ServerController.php
+++ b/app/Http/Controllers/ServerController.php
@@ -388,6 +388,20 @@ class ServerController extends Controller
         ]);
     }
 
+    /**
+     * Determine whether the user has reached the per-product server limit.
+     */
+    private function productLimitReached(User $user, Product $product): bool
+    {
+        if ($product->serverlimit == 0) {
+            return false;
+        }
+
+        $productCount = $user->servers()->where('product_id', $product->id)->count();
+
+        return $productCount >= $product->serverlimit;
+    }
+
     private function getUpgradeOptions(Server $server, array $serverInfo): \Illuminate\Database\Eloquent\Collection
     {
         $currentProduct = Product::find($server->product_id);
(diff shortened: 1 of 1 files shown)


MrWeez (changes requested):
The limit-check logic is duplicated three times. Same block (`count >= serverlimit && serverlimit != 0`) is copy-pasted across `getUpgradeOptions`, `upgrade()`, and `validateUpgrade()`. On top of that, the check in `upgrade()` is effectively duplicated by the one inside `validateUpgrade()`, which `upgrade()` calls right afterward meaning the same COUNT query hits the DB twice within a single request. Worth extracting into a single private method, and reuse it everywhere (see example bellow), leaving `upgrade()` only responsible for returning its specific error message.
```php
private function productLimitReached(User $user, Product $product): bool
```

MrWeez (approved):
Awesome!

PR #1358, opened 2026-07-12
Enforce per-node allocation limit when finding available nodes

### LINKED ISSUE #1354: Allocation Limit works per-location instead of per-node
### Pre-flight checklist

- [x] I have searched existing issues and this has not been reported before
- [x] I am using the latest available release

### Branch

main

### Bug Description

Allocation Limit must be treated as per-node option, but currently it prevents creation of new servers on all nodes in the same location

### Steps to Reproduce

1. Set "Allocation Limit" in Settings -> Server to n+1 amount where n is amount of servers on pterodactyl node 1
2. Add node 1 and node 2 from same location to product
3. Create server on this location, everything will work
3. Create another server, it will fail while only node 1 reached it Allocation Limit

### Expected Behavior

Server should be created on node 2

### CtrlPanel Logs

```Shell
https://ptero.co/RYdIYs
```

### Additional Info

Also i checked with only using Node2 on the products and the same issue still occurs

KroZen-Dev:
## Description

Enforce the per-node allocation limit when selecting an available node for
server provisioning. Previously, `findAvailableNode()` only rejected nodes
that lacked resources or had no free allocations, and
`findAvailableNodeWithAllocation()` only checked resources — neither checked
whether a node had already reached its per-node allocation limit. As a result,
nodes at their allocation limit could still be treated as available and only
fail later when trying to grab an allocation.

This adds an early guard in both methods: if
`nodeHasReachedAllocationLimit($node, $node->allocation_limit)` is true, the
node is rejected before any free-allocation lookup, making node selection more
accurate and avoiding wasted calls.

## Type of Change

- [x] Bug fix

## Testing

- Verified `findAvailableNode()` rejects a node once its per-node allocation
  limit is reached.
- Verified `findAvailableNodeWithAllocation()` skips nodes at their allocation
  limit and falls through to the next available node.
- Confirmed existing resource and free-allocation checks still behave as before.

## Checklist

- [x] My PR targets the `development` branch
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Code follows PSR-12
- [x] I have reviewed my own code
- [x] I have tested all affected functionality
- [x] No new warnings or errors introduced

### DIFF
--- a/app/Classes/PterodactylClient.php
+++ b/app/Classes/PterodactylClient.php
@@ -242,6 +242,42 @@ class PterodactylClient
     }
 
     /**
+     * Check whether a node has reached its allocation limit.
+     *
+     * The allocation limit is a per-node setting: it caps the number of
+     * allocations that may be in use on a single node. When the amount of
+     * assigned allocations on the node meets or exceeds the limit, no new
+     * servers can be deployed to that node (but other nodes in the same
+     * location remain eligible).
+     *
+     * @param  Node  $node
+     * @return bool
+     */
+    public function nodeHasReachedAllocationLimit(Node $node, int $nodeAllocationLimit = null): bool
+    {
+        // A limit of 0 means "no limit", so the node is never blocked.
+        $limit = $nodeAllocationLimit ?? $this->allocation_limit;
+        if ($limit <= 0) {
+            return false;
+        }
+
+        $response = self::getAllocations($node);
+
+        if (!isset($response['data']) || empty($response['data'])) {
+            return false;
+        }
+
+        $usedAllocations = 0;
+        foreach ($response['data'] as $allocation) {
+            if ($allocation['attributes']['assigned']) {
+                $usedAllocations++;
+            }
+        }
+
+        return $usedAllocations >= $limit;
+    }
+
+    /**
      * @param  Node  $node
      * @return array|mixed
      *
@@ -250,7 +286,7 @@ class PterodactylClient
     public function getAllocations(Node $node)
     {
         try {
-            $response = $this->application->get("application/nodes/{$node->id}/allocations?per_page={$this->allocation_limit}");
+            $response = $this->application->get("application/nodes/{$node->id}/allocations?per_page={$this->per_page_limit}");
         } catch (Exception $e) {
             throw self::getException($e->getMessage());
         }
--- a/app/Classes/PterodactylClient.php
+++ b/app/Classes/PterodactylClient.php
@@ -220,12 +220,15 @@ class PterodactylClient
     /**
      * @param  Node  $node
      * @return array|mixed|null
-     *
-     * @throws Exception
      */
     public function getFreeAllocations(Node $node)
     {
-        $response = $this->getAllocations($node);
+        try {
+            $response = $this->getAllocations($node);
+        } catch (\Exception $e) {
+            return [];
+        }
+
         $freeAllocations = [];
 
         if (isset($response['data'])) {
@@ -261,7 +264,11 @@ class PterodactylClient
             return false;
         }
 
-        $response = $this->getAllocations($node);
+        try {
+            $response = $this->getAllocations($node);
+        } catch (\Exception $e) {
+            return true;
+        }
 
         if (!isset($response['data']) || empty($response['data'])) {
             return false;
--- a/app/Classes/PterodactylClient.php
+++ b/app/Classes/PterodactylClient.php
@@ -225,7 +225,7 @@ class PterodactylClient
      */
     public function getFreeAllocations(Node $node)
     {
-        $response = self::getAllocations($node);
+        $response = $this->getAllocations($node);
         $freeAllocations = [];
 
         if (isset($response['data'])) {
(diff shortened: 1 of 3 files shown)


app/Classes/PterodactylClient.php
```diff
@@ -241,6 +241,42 @@ public function getFreeAllocations(Node $node)
...
+     * The allocation limit is a per-node setting: it caps the number of
+     * allocations that may be in use on a single node. When the amount of
+     * assigned allocations on the node meets or exceeds the limit, no new
+     * servers can be deployed to that node (but other nodes in the same
+     * location remain eligible).
+     *
+     * @param  Node  $node
+     * @return bool
+     */
+    public function nodeHasReachedAllocationLimit(Node $node, int $nodeAllocationLimit = null): bool
```
MrWeez: 1. The second parameter `$nodeAllocationLimit` is redundant since method already takes `Node $node`, and at all 4 call sites you're passing `$node->allocation_limit` i.e. a value from the same object that's already the first argument. Also, the `?? $this->allocation_limit` fallback is currently dead code, since no caller relies on the default. It'd be simpler and safer to drop the second parameter and just read `$node->allocation_limit` inside the method
2. `int $nodeAllocationLimit = null` - on PHP 8.1+ this triggers a deprecation notice (implicit nullable parameter). Needs to be `?int $nodeAllocationLimit = null`, or just remove the parameter per point 1

app/Classes/PterodactylClient.php
```diff
@@ -241,6 +241,42 @@ public function getFreeAllocations(Node $node)
...
+    {
+        // A limit of 0 means "no limit", so the node is never blocked.
+        $limit = $nodeAllocationLimit ?? $this->allocation_limit;
+        if ($limit <= 0) {
+            return false;
+        }
+
+        $response = self::getAllocations($node);
```
MrWeez: `getAllocations()` isn't static, but here it's called via `self::`. This works in PHP (calling a non-static method via `self::` from within an object context is effectively equivalent to `$this->`), but it's non-idiomatic, confusing to read, and could behave unexpectedly if the method is ever made abstract or overridden in a subclass. Change it to `$this->getAllocations($node)`.

app/Classes/PterodactylClient.php
```diff
@@ -241,6 +241,42 @@ public function getFreeAllocations(Node $node)
...
+        $usedAllocations = 0;
+        foreach ($response['data'] as $allocation) {
+            if ($allocation['attributes']['assigned']) {
+                $usedAllocations++;
+            }
+        }
+
+        return $usedAllocations >= $limit;
+    }
+
```
MrWeez: `getAllocations()` throws via `self::getException()` on API failure. If a single Pterodactyl node is temporarily unreachable, that exception will bubble straight out of the `reject()`/filter closure in `findAvailableNode()` and likely take down the entire node-selection process for every other (available) node in that location instead of just skipping the problematic one. Worth wrapping the `getAllocations` call in a try/catch, and on API failure treat the node as unavailable (`return true`)

MrWeez (changes requested):

app/Classes/PterodactylClient.php
```diff
@@ -253,15 +253,15 @@ public function getFreeAllocations(Node $node)
...
     {
         // A limit of 0 means "no limit", so the node is never blocked.
-        $limit = $nodeAllocationLimit ?? $this->allocation_limit;
+        $limit = $node->allocation_limit;
         if ($limit <= 0) {
             return false;
         }
 
-        $response = self::getAllocations($node);
+        $response = $this->getAllocations($node);
```
MrWeez: It would be better to wrap the `$response` with a try/catch block instead of wrapping every call site

```suggestion
try {
    $response = $this->getAllocations($node);
} catch (\Exception $e) {
    return true;
}
```

With that implementation, you can keep

```php
if ($this->pterodactyl->nodeHasReachedAllocationLimit($node)) {
    return true;
}
```

app/Http/Controllers/ServerController.php
```diff
@@ -573,12 +573,16 @@ private function findAvailableNode(string $locationId, Product $product): ?Node
...
-            if ($this->pterodactyl->nodeHasReachedAllocationLimit($node, $node->allocation_limit)) {
+            if ($this->pterodactyl->nodeHasReachedAllocationLimit($node)) {
+                return true;
+            }
+
+            try {
+                $freeAllocations = $this->pterodactyl->getFreeAllocations($node);
+            } catch (\Exception $e) {
                 return true;
             }
```
MrWeez: Because we wrapped `$response` inside `getFreeAllocations()` (see review comment for that function) you don't need to wrap call site here

app/Services/ServerCreationService.php
```diff
@@ -313,12 +313,16 @@ private function findAvailableNode(string $locationId, Product $product): ?Node
...
-            if ($this->pterodactylClient->nodeHasReachedAllocationLimit($node, $node->allocation_limit)) {
+            if ($this->pterodactylClient->nodeHasReachedAllocationLimit($node)) {
+                return true;
+            }
+
+            try {
+                $freeAllocations = $this->pterodactylClient->getFreeAllocations($node);
+            } catch (\Exception $e) {
                 return true;
             }
```
MrWeez: Same as `ServerController.php` line 580-584 (see review comment)

app/Classes/PterodactylClient.php
```diff
@@ -225,7 +225,7 @@ public function getFreeAllocationId(Node $node)
      */
     public function getFreeAllocations(Node $node)
     {
-        $response = self::getAllocations($node);
+        $response = $this->getAllocations($node);
```
MrWeez: It would be better to wrap the `$response` with a try/catch block instead of wrapping every call site

```suggestion
try {
    $response = $this->getAllocations($node);
} catch (\Exception $e) {
    return [];
}
```

With that implementation, you can keep

```php
$freeAllocations = $this->pterodactyl->getFreeAllocations($node);
return empty($freeAllocations);
```

MrWeez (changes requested):

MrWeez (approved):

MrWeez:
Good work!

